PR #37, opened 2016-05-16
A minor fix for PageInfo's endCursor retrieval (extra trailing space)

OlegZee:
Hi there! Thank you for a great library and here're mine couple bits to make even better.
The issue is minor while is took me couple hours to guess what's wrong (query returned null for endCursor).


Horusiath:
Great! Thanks for finding it, that was definitely hard to catch.

Issue #2, opened 2016-04-01
Fix parser


Horusiath:
Current parser suffers from some issues, where most notable is being fragile on the newline/whitespaces in front of the query.

The issue is recreated by _parser should parse kitchen sink_ test.

Known issues of the current parser:
- Problems with parsing queries having new lines in them, especially at the beginning of the query.
- Problems with parsing complex inputs as an arguments.

PR #42, opened 2016-06-01
Handling multiple input variables #2.10

OlegZee:
Another minor fix.


Horusiath:
Great! We're going to introduce a modified parser in the future, to handle complex object inputs and reduce parsing time.

Issue #46, opened 2016-06-21
Add support for complex input objects


Horusiath:
Right now we still have a rich set of tests that are failing because of lack of support for complex input arguments. This is a thing that needs to be solved in order to use any advanced mutations - like those required by Relay.js.

I guess, that after #2 we can finally proceed with this issue.

Links:
- [GraphQL specification of input objects](https://facebook.github.io/graphql/#sec-Input-Object-Values)

PR #47, opened 2016-06-24
Support for typed arguments

Horusiath:
/cc #46

This is a first approach to start supporting arguments set in inline GraphQL query. Limitations, that are present at this moment:
- If argument is complex `InputObject` it must have a concrete type associated with it (so you cannot bind `Define.InputObject` to things like maps yet).
- Argument values parsed from field are supplied to type instance through it's constructor, so type has to expose constructor with all fields that could be potentially used to pass data. F# records satisfy this rule by default.
- The only allowed collection type for input fields at the moment is `'t seq` - internally it's an F# list. This will be one of the hardest things to solve.

I've fixed some of the tests from VariableTests.fs - I'll take more of them next week.


Horusiath:
This should be working and ready to merge.

Issue #35, opened 2016-05-13
GraphQL client


Horusiath:
To give an ability to work with GraphQL not only on the server side, but also on the client and service-to-service, we should create a client for it. This is a very good place to introduce Type Provider, as we could generate the whole GraphQL API from simple introspection query.

Notes:
- We shouldn't use hard dependency on JSON serializer. GraphQL doesn't specify serialization format, which should be left to user (or plugable). Example: in the case of cross microservice calls, faster and more compact formats would be probably better than JSON.

Links:
- [GraphQL introspection query](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/blob/dev/src/FSharp.Data.GraphQL/Introspection.fs#L14), which by default returns a schema information.
- [Introspection specification](https://facebook.github.io/graphql/#sec-Introspection)
- Great [blog section](http://blog.mavnn.co.uk/blog/categories/typeprovider/) about type providers.

Horusiath:
Closing this now. Relay sample is working to some extend. Further work needs #2 to be fixed first.

Horusiath:
Closing this one: client library has been created. Further improvements will have their own issues.

PR #50, opened 2016-07-01
WIP: Client Type Provider

alfonsogarciacaro:
Here comes the promised PR for the Type Provider. At the end I'm trying a mixed approach: I read the Query objects from the schema and build methods based on that (with the query as a static argument so it can be checked at design time) so the query names and arguments are automatically added (the SelectionSet must still be passed as a string). @Horusiath please have a look and tell me what you think and which features we should focus on. I'm writing some ideas below.
- [ ] Improve query validation: right now it's just running `parse` on the query string. It should validate the name of the fields, etc, as it's done by [graphiql](https://github.com/graphql/graphiql)
- [ ] Support mutations
- [ ] Build types for queries according to the fields requested? Though according to what we talked this probably doesn't add much value for the work involved
- [ ] Anything else?

### DIFF
--- a/samples/client-provider/query.fsx
+++ b/samples/client-provider/query.fsx
@@ -1,0 +1,10 @@
+﻿#r "../../src/FSharp.Data.GraphQL.Client/bin/Debug/FSharp.Data.GraphQL.Client.dll"
+
+open FSharp.Data.GraphQL
+
+let [<Literal>] serverUrl = "http://localhost:8083"
+let [<Literal>] query = "{ hero(id: \"1000\") { name } }"
+
+type MyClient = GraphQLProvider<serverUrl>
+
+MyClient.Query<query>() |> Async.RunSynchronously |> printfn "%A"
--- a/samples/client-provider/query.fsx
+++ b/samples/client-provider/query.fsx
@@ -6,7 +6,7 @@ open System.Collections.Generic
 let [<Literal>] serverUrl = "http://localhost:8083"
 
 // The name and arguments of the query will be automatically set by the type provider
-let [<Literal>] queryFields = "{ id, name, friends { name } }"
+let [<Literal>] queryFields = "{ id, name, appearsIn, friends { name } }"
 
 type MyClient = GraphQLProvider<serverUrl>
 
@@ -23,6 +23,8 @@ match hero with
 | None -> ()
 | Some hero ->
     printfn "My hero is %A" hero.name
+    printfn "Appears in %O: %b" MyClient.Episode.Empire
+        (hero.appearsIn |> Array.exists ((=) MyClient.Episode.Empire))
     printfn "My hero's friends are:"
     hero.friends
     |> Array.choose (fun x -> x.name)
--- a/samples/client-provider/query.fsx
+++ b/samples/client-provider/query.fsx
@@ -1,10 +1,19 @@
 ﻿#r "../../src/FSharp.Data.GraphQL.Client/bin/Debug/FSharp.Data.GraphQL.Client.dll"
 
 open FSharp.Data.GraphQL
+open System.Collections.Generic
 
 let [<Literal>] serverUrl = "http://localhost:8083"
-let [<Literal>] query = "{ hero(id: \"1000\") { name } }"
+let [<Literal>] query = "{ hero(id: \"1000\") { id, name } }"
+
+let (?) (o: obj) (k: string) = (o :?> IDictionary<string,obj>).Item(k)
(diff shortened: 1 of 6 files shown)


samples/client-provider/query.fsx
```diff
@@ -0,0 +1,37 @@
...
+let [<Literal>] serverUrl = "http://localhost:8083"
+
+// The name and arguments of the query will be automatically set by the type provider
+let [<Literal>] queryFields = "{ id, name, appearsIn, friends { name } }"
+let [<Literal>] queryFieldsWithFragments = "{ ...data, friends { name } } fragment data on Human { id, name, appearsIn }"
+
+type MyClient = GraphQLProvider<serverUrl>
+
+let hero =
+    MyClient.QueryHero<queryFieldsWithFragments>("1000")
```
Horusiath: In this case you're creating statically typed top level query field. However I think it would be good to still leave the possibility of specifying it from string, like `{ hero("1000") { id, name, appearsIn, friends { name } } }`.

Reason for that is that user may define his own top level query and use multiple queries defined in schema as part of single execution. Take a look at [MutationTests](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/blob/dev/tests/FSharp.Data.GraphQL.Tests/MutationTests.fs#L77) for examples.

alfonsogarciacaro: You're right, there should also be a generic `Query` method, I'll add it :+1: Thanks!

samples/client-provider/query.fsx
```diff
@@ -0,0 +1,37 @@
...
+    MyClient.QueryDroid<queryFields>("2000")
+    |> Async.RunSynchronously
+
+// Result is an option type
+match hero with
+| None -> ()
+| Some hero ->
+    printfn "My hero is %A" hero.name
+    printfn "Appears in %O: %b" MyClient.Episode.Empire
+        (hero.appearsIn |> Array.exists ((=) MyClient.Episode.Empire))
```
Horusiath: `MyClient.Episode.Empire` this looks problematic, as generated types resides in the same scope as top level methods. What if we had `Episode` type defined and `episode` field inside schema's query?

Sql provider solves that by adding extra `Types` scope i.e. `MyClient.Types.Episode`.

Horusiath:
I think, that query string validation can come later. Doesn't have to be here right away. 

Other remarks:
- We could separate queries from mutations and subscriptions by creating separate scopes for them i.e. Instead of using convention `MyClient.HeroQuery` (_what if that field name will be "heroQuery"? "HeroQueryQuery"?_), maybe use `MyClient.Queries.Hero`, `MyClient.Mutations.Hero` and so on?

Beside that, everything looks cool :+1: Please rebase with dev (remember to update client.fsx example) and we're ready to merge.

/cc #35

samples/client-provider/query.fsx
alfonsogarciacaro: This is easy to do and will be safer indeed. BTW, I've a question about Enums: currently I'm using the value names as the possible values as I cannot find a `Value` field in `IntrospectionEnumVal`. According to [this](http://graphql.org/docs/api-reference-type-system/#graphqlenumtype), _GraphQL serializes Enum values as strings, however internally Enums can be represented by any kind of type_. Is this _internal representation_ only happening in the server and enum values are always sent to the client represented with their string names.. In other words, it's [my current approach](https://github.com/alfonsogarciacaro/FSharp.Data.GraphQL/blob/client-provider/src/FSharp.Data.GraphQL.Client/IntrospectionParser.fs#L145-L148) right? Note, for example, [I had to modify the case of enum values](https://github.com/alfonsogarciacaro/FSharp.Data.GraphQL/commit/4f0d5a017b610768a2635c164d8143c0d1685e6c#diff-e55cde849b1089690809ecfacdc8847fL92) in the `server.fsx` example to make it work right.

Horusiath: I think we cannot make enums working 1-1 with their server representation. As mentioned, for the client they are represented as strings - that's all we can guarantee. Actually we cannot even be sure that the server is written in language that supports enums ;)

alfonsogarciacaro:
Thanks for merging and sorry for the delay, there was a minor error in the server script that troubled me a bit. I implemented your comments. I'm having a first look now at validation and I'm also trying to implement the type provider with Fable so it's easier to use from the client side.

Cheers!

Horusiath:
Don't worry. Everything is fine. Regarding Fable, great idea! :)

PR #54, opened 2016-07-08
Adapt type provider to Fable

alfonsogarciacaro:
First attempt to adapt the type provider to Fable, though is not working yet. I'm currently on OSX and I'm having problems with the Hopac dependency. Could it be possible to remove the Hopac dependency from the `FSharp.Data.GraphQL` components the type provider relies on?


Horusiath:
I think we could move some functionalities shared between client and server to some core library. If you want, you may try that by yourself, or leave it to me.

alfonsogarciacaro:
Perfect, if you're OK with that I'll give it a try :+1: I'll also check it with with Windows anyways, maybe it's a Mono thing. Cheers!

alfonsogarciacaro:
I've removed the `FSharp.Data.GraphQL` dependency from the client project. The FParsec dependency remains but Hopac is gone too :+1: For simplicity I just added links to the necessary files from `FSharp.Data.GraphQL.Client`, but please tell me if you prefer to move the shared files to an independent project. For now, as I don't need the methods using `Hopac.Job`, I've replaced them with [placeholder values](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/files#diff-c587c63b2d69581bd410f505a83de809R9) but this should be fixed in the future.

The provider is working with `fable-compiler@next` and the generated JS can be run with node. This can be merged if you want.

Horusiath:
I think, we should move shared parts to a separate project. How much of these would be? I'm sure about type system and introspection. What about the rest?

alfonsogarciacaro:
That'll be the files in the PR [linked from the client project](https://github.com/alfonsogarciacaro/FSharp.Data.GraphQL/blob/768b6d3f8385b0803ad533835b5b751120dfd9fd/src/FSharp.Data.GraphQL.Client/FSharp.Data.GraphQL.Client.fsproj#L164-L178).

If we want to remove the dependency to Hopac from the client project, the trickiest part is the references to `Job` in `TypeSystem.fs` (like [this one](https://github.com/alfonsogarciacaro/FSharp.Data.GraphQL/blob/768b6d3f8385b0803ad533835b5b751120dfd9fd/src/FSharp.Data.GraphQL/TypeSystem.fs#L229)). We could create an interface for them but maybe it's easier to just use `Async` as I think it was before, what do you think?

Horusiath:
Problem with Async is that it's slow, and even in combination with Hopac it will introduce a speed penalty. We could try to go back for it (only for field resolution type signature) and then try to optimize again from that point.

alfonsogarciacaro:
No worries, I'll use an interface to abstract `Job` then :+1: I'll explore the code in `FSharp.Data.GraphQL` better to see how can we _inject_ the code to deal with the jobs.

alfonsogarciacaro:
I have moved the shared files to a new project which is compiled as a PCL and, in principle, should be also compatible with dotnetcore. I had to change some code dealing with reflection for that (for example, [here](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/commits/6e389c232058fe07e1a44027df6e62d2bc7e6147#diff-0e76f155fa9dabcf49857277311dce49L40) and [here](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/commits/6e389c232058fe07e1a44027df6e62d2bc7e6147#diff-81be4a28be52354b20cb22fedd8f2770L1245)) besides removing the `SerializableAttribute` of some types.

As discussed, I tried to abstract `Job` so the other projects (server and client) can implement it however they like. Ideally the other project should pass the code to deal with jobs, but unfortunately some of the Job methods are used in [record constructors](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/blob/dev/src/FSharp.Data.GraphQL/TypeSystem.fs#L506) which don't accept additional arguments. To overcome this, I've implemented a very simple [dependency injection](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/commits/6e389c232058fe07e1a44027df6e62d2bc7e6147#diff-06b89e2ab6211f077cc715bebafaa6caR6) (check how it's used [here](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/commits/6e389c232058fe07e1a44027df6e62d2bc7e6147#diff-81be4a28be52354b20cb22fedd8f2770R12) and [here](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/commits/6e389c232058fe07e1a44027df6e62d2bc7e6147#diff-401652ed3974ddf1c721d70f614a5b7dR21)). But I'm still not very happy with it. Please have a look and tell me what you think: whether we should go in this direction, modify the constructor of `FieldDefinition` instead, replace `Job` with `Task` or something else.

(BTW, the server sample is working right now, but I still need to fix the tests.)

Also, if we move most of the shared files to a Shared project, it may make sense to rename `FSharp.Data.GraphQL` to `FSharp.Data.GraphQL.Server`, what do you think?

Horusiath:
I think we could go back to plain old `Async<>` for now. I'm thinking about some more advanced solution, where we could use `Define.Field` and `Define.AsyncField` to register a dedicated data structures (something like `type Resolve = Sync of ResolveFn | Async of ResolveFnAsync`) and then utilize our compilation phase to create sync/async executors for them. But that is an idea for later.

alfonsogarciacaro:
Perfect, then I'll just Async, this will be much simpler :+1:

alfonsogarciacaro:
As discussed, I eliminated the Job abstraction and now the shared project just uses Async. However, `FSharp.Data.GraphQL` keeps having the Hopac reference.

Also, I noticed that many tests require that type properties are searched for ignoring the case, which is not possible by default in netcore's `TypeInfo.GetDeclaredProperty` so I've added an [extension](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/54/files#diff-5deec133de63662669476cf927682653R10) for that. Note the extension only changes the case of the first letter which should be faster than listing all properties and comparing one by one with ignoring the case, but please tell me if you prefer this latter method.

Tests are passing but the build is failing in the `GenerateReferenceDocs` step. I've also rebased the branch with the latest code in `dev` so if you agree, I think we can merge this so it's easier to fix the build with the new project structure.

alfonsogarciacaro:
The PR is finally building in AppVeyor. I'm a bit puzzled about Travis because the project does build on my OSX with Mono 4.4.1.

kristianmandrup:
@alfonsogarciacaro Any resources on how to integrate this lib with fable? I'd like to use F#, Fable and GraphQL in combination. Any advice or suggestions would be most welcome :)

Ah, perhaps this one: https://fsprojects.github.io/FSharp.Data.GraphQL/client-side.html

But looks very brief. Any sample projects out there?

alfonsogarciacaro:
This was done 2 years with one of the very first versions of Fable. However I didn't have a good experience with GraphQL and haven't used it in my projects since. So unfortunately I don't have any plans to work on updating this type provider or anything GraphQL related. That would have to be done by someone with more GraphQL expertise than me, sorry.

kristianmandrup:
I'll give it a go then as one of my first contributions ;)

PR #55, opened 2016-07-09
Adds support for Subscription

mrinaldi:
This PR introduces experimental support for subscriptions.

The specs for subscriptions were not released yet, but there are some info around it
- http://graphql.org/blog/subscriptions-in-graphql-and-relay/
- https://gist.github.com/OlegIlyenko/a5a9ab1b000ba0b5b1ad

### DIFF
--- a/src/FSharp.Data.GraphQL/Execution.fs
+++ b/src/FSharp.Data.GraphQL/Execution.fs
@@ -490,6 +490,11 @@ type Define with
                          |> Array.filter (fun f -> f.Name = name)
                          |> Array.fold (fun _ f -> f.SelectionSet) []
 
+                     let label =
+                         ctx.Fields
+                         |> Array.filter (fun f -> f.Name = name)
+                         |> Array.fold (fun _ f -> f.AliasOrName) name
+
                      let execute x =
                         collectFields ctx.ExecutionContext typedef set (ref [])
                         |> executeFields ctx.ExecutionContext typedef x
@@ -498,7 +503,7 @@ type Define with
                      |> Async.Global.ofJob
                      |> Async.RunSynchronously
                      |> Observable.map (fun x ->
-                        [name,
+                        [label,
                             execute x
                             |> Async.Global.ofJob
                             |> Async.RunSynchronously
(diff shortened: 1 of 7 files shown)


src/FSharp.Data.GraphQL/Execution.fs
```diff
@@ -470,4 +474,49 @@ let internal execute (schema: #ISchema) doc operationName variables root errors
...
+    /// Single subscription defined inside either object types or interfaces 
+    static member Subscription(name : string, typedef : ObjectDef<'Res>,
+                               subscribe : ResolveFieldContext -> 'Val -> IObservable<NameValueLookup> -> unit,
+                               ?resolve: ResolveFieldContext -> 'Val -> IObservable<'Res>,
+                               ?description: string, ?args : InputFieldDef list, ?deprecationReason : string) : FieldDef<'Val> =
+        upcast { Name = name
+                 Description = description
+                 Type = typedef
+                 Resolve = fun ctx value -> job {
+                     let set =
```
Horusiath, line 490: I think, that this kind of logic should not reside here. If necessary, it should be probably computed at the moment when Schema is created (just like `ExecuteInput` and `IntrospectionSchema`).

src/FSharp.Data.GraphQL/Execution.fs
```diff
@@ -470,4 +474,49 @@ let internal execute (schema: #ISchema) doc operationName variables root errors
...
 TypeMetaFieldDef.Execute <- compileField Unchecked.defaultof<TypeDef -> ObjectDef[]> TypeMetaFieldDef
-TypeNameMetaFieldDef.Execute <- compileField Unchecked.defaultof<TypeDef -> ObjectDef[]> TypeNameMetaFieldDef
\ No newline at end of file
+TypeNameMetaFieldDef.Execute <- compileField Unchecked.defaultof<TypeDef -> ObjectDef[]> TypeNameMetaFieldDef
+
+
+type Define with
+    /// Single subscription defined inside either object types or interfaces 
+    static member Subscription(name : string, typedef : ObjectDef<'Res>,
+                               subscribe : ResolveFieldContext -> 'Val -> IObservable<NameValueLookup> -> unit,
```
Horusiath, line 483: I see two problems here:
1. We can define only single publisher / single subscriber relationship. Multiple subscribers are not allowed.
2. All of the methods existing in `Define` type are reflecting native GraphQL type system definitions. There is no such type as subscription. It's artificial construct from our side and there's a risk that it won't fit GraphQL standard in the future. Therefore subscription specific logic should be done totally in the underlying execution engine.

PR #62, opened 2016-07-28
WIP: Add Nuget build targets

alfonsogarciacaro:
I've added two new build targets to publish the server and client packages on Nuget (I've also added the `.Server` suffix to the server project). At the moment, the targets only work if run specifically (e.g. `build PublishServer`) on a machine where Bazinga's Nuget API key has been set.

Things to consider:
- [ ] Make the publish automatic on the release server?
- [ ] Add the Shared project as an independent package?
- [ ] Remove the old nuget targets? (Nuget, PublishNuget)

Note: A beta version of both packages has already been released to Nuget.


alfonsogarciacaro:
I added a new target to publish the Type Provider version for Fable through npm.

tools/Nuget/template.nuspec
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Horusiath: Aren't we going to use Paket templating instead of NuGet nuspec files?

alfonsogarciacaro: Anyone it's ok for me, but when I checked other repos and FAKE documentation that's how it was done: http://fsharp.github.io/FAKE/create-nuget-package.html

Horusiath: I've used Paket templating in one of my projects - [here](https://github.com/Horusiath/Akkling/blob/master/src/Akkling/Akkling.fsproj.paket.template) is the example paket.template file, and [here](https://github.com/Horusiath/Akkling/blob/master/build.fsx#L144) you can see, how to use Paket commands from within the FAKE build scripts.

alfonsogarciacaro: Perfect, thanks! I'll change the targets then :+1:

PR #63, opened 2016-08-05
Todo task sample with GraphQL

alfonsogarciacaro:


alfonsogarciacaro:
I still need to add a UI but this simple example can already be merged, as it includes improvements for the client type provider to support mutation methods (the build seems to be failing again because of the problem when generating documentation though).
Note the server script is already using the prereleased version of FSharp.Data.GraphQL from Nuget.

PR #68, opened 2016-08-15
Typed queries

alfonsogarciacaro:
This is the first proof of concept for typed queries. In its current form the GraphQL type provider just uses a literal string to specify which fields the user wants to receive from the server. For example:

``` fsharp
let [<Literal>] queryFields = "{ id, name, appearsIn, friends { name } }"
let [<Literal>] queryFieldsWithFragments = "{ ...data, friends { name } } fragment data on Human { id, name, appearsIn }"

type MyClient = GraphQLProvider<serverUrl>

async {
    let! hero = MyClient.Queries.Hero<queryFieldsWithFragments>("1000")
    ...
}
```

There're a couple of advantages of these approach but not many:
- It's not necessary to write the name and arguments of the query (here: `{ hero(id:"1000") ... }`) as this is determined by the provided method.
- The string is checked a compile time (that's why the string must be a literal passed a a static argument) so the code won't compile if the query is malformed.

However, these advantages don't seem to be enough to justify a type provider, so it's worth exploring some ways to make the projection in a type-safe manner with intellisense. With this goal, first I tried to implement a custom query expression builder, as in [this article](http://tomasp.net/blog/2015/query-translation/), which could look like this:

``` fsharp
query { 
  for h in MyClient.Queries.Hero do
  where (h.id = 10)
  select (h.name, h.appearsIn) }
```

However, I discarded this approach because of two reasons:
- In query expressions, the user decides the query source. In the GraphQL the source is defined by the provided query method, which also defines the arguments valid for that query.
- The operations in GraphQL (filtering, grouping...) are defined by the arguments, so we only need a simple projection for the result and building a whole query expression seems overkill for that.

The projection may be expressed with a simple lambda and that's what I'm trying to do at the moment: 

``` fsharp
MyClient.Queries.Task(taskId,
    <@@ fun (t: MyClient.Types.Task) -> t.description, t.completed @@>)
```

> Ideally a simple lambda should be enough instead of a explicit quotation. 

For testing purposes, I'm just removing the old methods which accepted a literal string as a static argument. We need to decide if we want to have both solutions side by side, or we focus on this new approach to avoid confusion (and maybe add a simple method to easily send string queries at runtime in case of need).

### DIFF
--- a/samples/client-provider/query.fsx
+++ b/samples/client-provider/query.fsx
@@ -10,22 +10,13 @@ Fable.Core.JsInterop.importAll<unit> "isomorphic-fetch"
 
 let [<Literal>] serverUrl = "http://localhost:8083"
 
-// The name and arguments of the query will be automatically set by the type provider
-let [<Literal>] queryFields = "{ id, name, appearsIn, friends { name } }"
-let [<Literal>] queryFieldsWithFragments = "{ ...data, friends { name } } fragment data on Human { id, name, appearsIn }"
-
 type MyClient = GraphQLProvider<serverUrl>
-
-// let droid =
-//     MyClient.Queries.Droid<queryFields>("2000")
-//     |> Async.RunSynchronously
 
 async {
     let! hero = MyClient.Queries.Hero("1000", fun c ->
                 Fields(
                     c.name,
-                    // c.appearsIn,
-                    Selection(c.friends, fun f -> Fields(f.[0].name)),
+                    Selection(c.friends, fun f -> Fields(f.name)),
                     On<MyClient.Types.Human>("Human", fun h -> Fields(h.appearsIn))
                 ))
     match hero with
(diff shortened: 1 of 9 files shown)


alfonsogarciacaro:
I managed to make the provided methods accept lambdas instead of explicit quotations. Now I've to see if I can remove the need for the `upcast`:

``` fsharp
MyClient.Queries.Task(taskId, fun t -> upcast (t.description, t.completed))
```

alfonsogarciacaro:
I haven't managed to use a generic argument (`Task->'T`) for the provided method, so I've created a dummy type to wrap the fields. It looks a bit more elegant than casting to `obj`.

``` fsharp
MyClient.Queries.Task(taskId, fun t -> Fields(t.description, t.completed))
```

alfonsogarciacaro:
Ok, the PR should be ready to merge now. As a follow-up of the Slack discussion, I tried to generate actual types but this proved to be complicated as there're many nested erased types being created at the moment and I also got strange errors on Mono [like this one](https://github.com/mono/mono/blob/master/mcs/class/corlib/System.Reflection.Emit/SignatureHelper.cs#L388), so now I'm solving the issue of passing the name of the type in inline fragments with type condition (`...on Human { name }`) by generating a static method `On` for every type in the GraphQL (atm I'm only modifying [GraphQL Object types](https://github.com/bazingatechnologies/FSharp.Data.GraphQL/pull/68/files#diff-f42367a0558d5b6edece1dec4bb356acR192)).

Right now, queries like the following are possible:

``` fsharp
open FSharp.Data.GraphQL
open System.Collections.Generic

type MyClient = GraphQLProvider<"http://localhost:8083">

async {
    let! hero = MyClient.Queries.Hero("1000", fun c ->
                Fields(
                    c.name,
                    Selection(c.friends, fun f -> Fields(f.name)),
                    MyClient.Types.Human.On(fun h -> Fields(h.appearsIn))
                ))
    match hero with
    | None -> ()
    | Some hero ->
        printfn "My hero is %A" hero.name
        printfn "Appears in %O: %b" MyClient.Types.Episode.Empire
            (hero.appearsIn |> Array.exists ((=) MyClient.Types.Episode.Empire))
        printfn "My hero's friends are:"
        hero.friends
        |> Array.choose (fun x -> x.name)
        |> Array.iter (printfn "- %s")
}
|> Async.StartImmediate
```

Which translates to:

``` graphql
query Query{
  hero(id: "1000") {
    name,
    friends {
      name
    },
    ...on Human {
       appearsIn
    }
  }
}
```

Please tell me if you have any comment or suggestion.

PR #69, opened 2016-08-19
Fix build and compile shared project for netcore

alfonsogarciacaro:
I’m trying to compile the project with netcore. I’ve managed to do it for the Shared project but I’m having problems with FSharp.Data.GraphQL.Client, mainly because of the type provider starter pack. I’m trying to modify it to make it compilable with netcore but I’m not sure yet if it’s worth the effort or not.


alfonsogarciacaro:
This PR can already be merged. It doesn't fulfil its original purpose (compiling all projects for netcore) but it lays the foundation for that so we can try again when the Hopac dependency is removed from the project and when we get an answer about type providers with netcore [here](https://github.com/fsprojects/FSharp.TypeProviders.StarterPack/issues/80#issuecomment-241479154). 

Besides that, this PR fixes build in Travis (and delays doc generation in AppVeyor until we fix it) by adding FSharp.Core to the projects as a Nuget package as recommended in this [SO question](http://stackoverflow.com/questions/31866034/why-does-my-f-project-build-locally-but-fail-on-travis).

PR #71, opened 2016-08-23
[WIP] Fix Fable conversion in Type Provider

alfonsogarciacaro:
Not working yet. I’m facing a problem because right now the provider is interpreting the quotation at runtime, which is not feasible yet in Fable. If possible, quotations should be resolved at compile time (static arguments maybe?).

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -399,12 +399,19 @@ Target "Release" (fun _ ->
     |> Async.RunSynchronously
 )
 
-Target "AdHocBuild" (fun _ ->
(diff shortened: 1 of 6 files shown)


nojaf:
@alfonsogarciacaro would this work with Fable 2? Any ideas?

alfonsogarciacaro:
Buf, not sure, it's been a while since I haven't touched GraphQL and totally forgot about this PR! 😅 Unfortunately, I don't have time right now to get back on this so I'll close this PR. I could give support to someone trying to retake it, but I'd recommend starting with a simpler TP to see how they work in Fable 2 (they should) and whether we need to implement anything to better support them

PR #74, opened 2016-08-26
fixed field alias parsing in case of no whitespace after colon character

OlegZee:
This is important issue because Relay client creates aliases without whitespace. E.g. 

```
query Routes {
          me {...F0  }
        }
        fragment F0 on Me {
          _channel2sNmIg:channel(key:\"sales\") {
              name,
              items {
                   id,
                   name,
                   author {login},
                   _meta {updatedAt, id}
              }
          }
        }
```


Horusiath:
Nice, you've catch that. Thanks! :)

PR #108, opened 2017-04-26
small typo fix

DmitryBatalov:
The word "edito" -> "editor". It is not easy to locate what was changed in diff view.


Horusiath:
Thanks :)

PR #110, opened 2017-05-05
Add experimental GraphQL subscription support

TOTBWF:
Several major graphql clients have begun to add experimental support for subscriptions, as detailed [here](http://graphql.org/blog/subscriptions-in-graphql-and-relay/). This pull request acts as a rough draft for adding support in this library. There are a few warts and kinks that may need working out, but as of now the system implements:

- Adding subscriptions to the schema
- Firing subscription events from inside of resolvers, as well as directly from the executor
- A flexible transport layer
- Support for synchronization across multiple machines (very experimental, and open for discussion with regards to the implementation)

**Example Usage**
Schema Definition
```fsharp
let Subscription = 
    Define.SubscriptionObject<Root>(
        name = "Subscription",
        fields = [
            Define.SubscriptionField(
                "watchMoon",
                RootType,
                PlanetType,
                "Watches to see if a planet is a moon",
                [ Define.Input("id", String) ],
                // Callback to be triggered on event fire
                (fun ctx r d -> printfn "Subscription with dictionary %A" d),
                // Filter function, used to determine if we care about an event
                (fun ctx r p -> ctx.Arg("id") = p.Id))])
```

Event Firing
```fsharp

    Define.Object<Root>(
        name = "Mutation",
        fields = [
            Define.Field(
                "setMoon", 
                Nullable PlanetType, 
                "Sets a moon status", 
                [ Define.Input("id", String); Define.Input("ismoon", Boolean) ], 
                fun ctx _ -> 
                    getPlanet(ctx.Arg("id")) 
                    |> Option.map (fun x -> 
                        x.SetMoon(Some(ctx.Arg("ismoon")))
                        // Subscription Handler is kept in the resolver context
                        ctx.SubscriptionHandler.FireEvent PlanetType x
```


Horusiath:
I can review it, but I think it was better idea to actually make a draft on paper first, before starting to implement things. Could you write some more about how have you implemented this one? Few of the questions that come to my mind at first:

1. Why does SubscriptionField even need to have RootType defined?
2. Why does subscription handler exists on the context level? That doesn't feel extensible at this point. Personally I was thinking about making an extra setting in server options in form of subscribe/unsubscribe functions that would be called once a client will try to send subscription query. The way how they would be handled, could rest in a hands of the user. But I think I'll write about this in separate issue.
3. I'm also not sure about a separate filter function in the field. This could be also detail of implementing library, because it would define a transport mechanism for published event, and because of that it would know what is the most optimal way merge/broadcast/filter them.

TOTBWF:
The basic implementation works in 2 phases: registration of the subscription fields in the schema at schema compilation time, and context application at runtime.
 
In the registration step, we insert the callback and filter functions into the subscriptionHandler. 


The reason for the filter function is that we may need to filter on values that are either computed, or available in the root type. For example, if we wanted to subscribe to all new messages coming to a given user, we could use the subscription:
```
subscription {
    watchMyMessages {
        body
        ...
    }
}
```
Note that this has no dependence on the arguments passed in, but obviously needs filtering in some way. 

This also ties into why we need the RootType (known as context in other implementations) as part of our definition: We need to know what data is available to the resolver, and how exactly we can represent that. This also ties into why we need the RootType (known as context in other implementations) as part of our definition: We need to know what data is available to the resolver, and how exactly we can represent that. On top of that, due to the way that some of the code is structured, we cannot remove the TypeDef field of SubscriptionFieldDef without screwing with a large portion of the code that deals with the typing of FieldDefs and ObjectDefs. This is something that could be looked at in the future, but I did not make it a priority when making this prototype.

When we receive a subscription, we resolve variables, fragments, etc. and feed it into our subscriptionHandler. The handler itself is created with the executor, and is threaded into the execution. It is also provided with a backing plane to manage the synchronization of subscriptions. In this example, this defaults to a concurrentDictionary, but other backing planes (databases, queues, etc) could be used. Adding it to the context is in no way final, it just felt like the most natural way for a user to fire events from a resolver. Hooks do exist in the executor to directly fire events outside of the context of a resolver.

When an event is fired, the subscriptionHandler dispatches the event to all interested subscriptions, who in turn resolve the fields that they requested as part of the query by using the same mechanisms as the resolvers for queries. It then applies the resolved dictionary to the callback, which the user would then feed to their desired transport mechanism.

If you have any further questions, please feel free to ask 🙂

Horusiath:
Could you rebase this one in order to merge?

TOTBWF:
Rebased, though I do think there is a discussion to be had about the api you proposed in #111

TOTBWF:
I've reworked the implementation to be compatible with the @defer PR @ @ #117, and reworked the api as well, so I will be closing this PR.

PR #112, opened 2017-05-12
Relay Modern's relay-compiler requires the connection to be optional.

lxwbr:
Relay Modern introduces a relay-compiler which yields an error if fields having a Connection type are not nullable. The nullable property will now be enforced by `ConnectionOf`.


Horusiath:
@zunder why did you closed that PR? Also it would be great if you could link the related change from Relay docs or something. It'd be a great help.

lxwbr:
@Horusiath I decided to not enforce then nullable property of a connection for now as the [Relay cursor connections specification](https://facebook.github.io/relay/graphql/connections.htm) does not explicitly mention it. However, the [relay-compiler](https://facebook.github.io/relay/docs/relay-compiler.html) is yielding an error in the case the connection is defined to be not-nullable.

What I decided for now is to fulfill the requirement in the definition part on the server side ([see server.fsx](https://github.com/zunder/FSharp.Data.GraphQL/blob/ec704b6d648f36638e7327963a26bc5ee78690f5/samples/relay-modern-starter-kit/server.fsx#L53)) and making a [new PR](https://github.com/fsprojects/FSharp.Data.GraphQL/pull/113) containing a new example using Relay Modern.

lxwbr:
@Horusiath as mentioned before, this was a [relay-compiler 1.0.0-rc.3](https://www.npmjs.com/package/relay-compiler) restriction, which is not existing in the current version of `relay-compiler`.

Issue #114, opened 2017-05-16
Merging netcore and msbuild2003 build files


TOTBWF:
### Description

With the release of Mono 5, netcore-style fsproj build files are supported by the packaged msbuild, meaning that all of the major build tools now use the same format. Should we consider merging the netcore and msbuild2003 build files into the unified format?

Horusiath:
Sure. It would be great :)

PR #116, opened 2017-05-18
Migrate buildfiles to new netcore style

TOTBWF:
See Issue #114. Note that there is a bug on OSX and Linux where the dotnet cli tools don't actually look for msbuild in your path, so you have to specify the target using the -f flag


TOTBWF:
It seems travis is still using xbuild even though it has been deprecated... I'll whip up a PR for them

Issue #102, opened 2017-03-02
[WIP] Uniform way for working with resolver results


Horusiath:
In one of [my blog posts](http://bartoszsypytkowski.com/c-a-different-look-at-service-design/) I've shown a way - inspired by Twitter's Finagle - of applying additional logic to service calls. I think we could apply something similar here. It's a proposal and needs some more polishing. 

Given a resolver function like:

```fsharp
type Resolve<'In, 'Out> = ResolveFieldContext -> 'In -> AsyncVal<'Out>
```
we could define a filter for it:

```fsharp
type Filter<'In1, 'In2, 'Out2, 'Out1> = 
    Emitter<'Out1> -> Resolve<'In1, 'Out2> -> ResolveFieldContext -> 'In1 -> AsyncVal<unit>
type Emitter<'Out> = 'Out -> unit
```

What is a filter? It's a wrapper around a resolver, that may apply some additional logic to it. It may (or may not) use emitter to nest value returned from resolver into result set set to user. What we could do with that?:

- It may be used as a general-purpose mechanism to implement directives. Currently we support only two of them (`@skip` and `@include`) and they are handled as special conditions. With filters we could implement directive logic as a filter that will be conditionally compiled into resolution pipeline.
- With emitters we could implement sequential results i.e. [@defer](https://dev-blog.apollodata.com/new-features-in-graphql-batch-defer-stream-live-and-subscribe-7585d0c28b07#8188) directive.

Some example to visualize:

```fsharp
// filter that could be used for @skip directive
let skipDirective emit resolve ctx input = asyncVal {
    let cond = ctx.Arg "if" 
    if not cond 
    then 
        // resolve a value if @skip(if: true)
        let! output = resolve ctx input
        emit output  // emit output to corresponding key-value result set
}
```

Horusiath:
## Emitters

Another idea regarding emitters. Right now we produce result by creating an object graph and serializing it into response stream. What if we provide some abstraction in form of an API (emitter), that will allow us to push data directly into response stream (ignoring object graph construction)? 

How this could work (it's not complete interface): 

```fsharp
type Emitter =
    interface 
        // basic value emitters
        abstract member Emit: bool -> unit
        abstract member Emit: int -> unit
        abstract member Emit: double -> unit
        abstract member Emit: string -> unit
        abstract member EmitNull: unit -> unit
        // list delimeters
        abstract member BeginList: unit -> unit
        abstract member EndList: unit -> unit
[...]

PR #117, opened 2017-05-23
Experimental @defer support

TOTBWF:
This PR adds support for the defer directive. Documentation is pretty sparse on it, and there are a couple of behaviours I'd like to work out with everyone (Nested defers, what to do with errors, etc). 

Implementing this necessitated a rewrite of our resolver execution code, because in our old model, our field selection would only give us the leaves of the result tree, and a deferred value could rely on a node.


TOTBWF:
Keep in mind that we are still waiting on https://github.com/travis-ci/travis-build/pull/1060 and https://github.com/fsprojects/FSharp.Control.Reactive/pull/93 to go through

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -368,13 +439,74 @@ let internal executePlan (ctx: ExecutionContext) (plan: ExecutionPlan) (objdef:
...
+                            let! tree' = treeToDict tree
+                            return NameValueLookup.ofList["data", upcast tree'; "path", upcast path]
+                        }) >> AsyncVal.collectParallel))
+                |> AsyncVal.collectParallel
+                |> AsyncVal.map(Array.fold (Array.append) [||])))
+            |> AsyncVal.collectParallel
+            |> AsyncVal.map(Array.fold (Array.append) [||])
+            |> AsyncVal.toAsync
+            |> Observable.ofAsync
+            |> Observable.bind(Observable.ofSeq)
```
Horusiath: If I read it right, you're basically load a whole collection to be deferred into memory, and `@defer` is just used to return it in chunks as `IObservable<>`? If it's true, I don't see any value added by defer directive.

paket.dependencies
```diff
@@ -1,9 +1,10 @@
 source https://nuget.org/api/v2
-nuget BenchmarkDotNet
-nuget BenchmarkDotNet.Diagnostics.Windows
+nuget BenchmarkDotNet 
+nuget BenchmarkDotNet.Diagnostics.Windows 
 nuget FParsec
 nuget FSCheck
 nuget FSharp.Data.TypeProviders
+nuget FSharp.Control.Reactive
```
Horusiath: Why do we need it for? `IObservable<>` is part of mscorlib.

Horusiath (commented):
We would need some benchmarks for that one. I'm also concerned about the implementation itself, as it looks like trying to prefetch all data upfront with the major data resolution flow, then serving it as Observable, while IMHO it should:

1. Continuously stream data as it's taken from the resolver (so resolver func returns an observable in this case, and maybe new custom type def). 
2. Fetch data aside using async, regardless of how type def has defined resolver - so in this case, deferred fields will always be resolved in parallel to the main resolution flow.

Another thing is that, we should try to bring a more general solution (I've started a discussion on that in #102). Define the requirements and API necessary to aim even such possibilities as deferred fields. But this is something that can eventually be done later and doesn't block this PR from merge.

paket.dependencies
TOTBWF: We need it for some of the transformations from other data types to observables, as well as the `bind` operation on observable

src/FSharp.Data.GraphQL.Server/ObservableExtensions.fs
```diff
@@ -0,0 +1,19 @@
+namespace FSharp.Data.GraphQL
+
+open System
+open System.Reactive
+open System.Reactive.Linq
+
+/// Extension methods to observable, used in place of Fsharp.Control.Reactive
+module Observable =
```
johnberzy-bazinga: We should probably mark this as internal.

Horusiath:
@TOTBWF What is the status on this one?

johnberzy-bazinga:
@Horusiath I think @TOTBWF is just waiting on travis-ci/travis-build#1060. If there is no movement on that PR in the next couple of days we can probably workaround it in the build script, or alternatively, just merge as-is.

TOTBWF:
This should be more or less good to go, there are a few little spots that display some odd behavior sometimes (nested fragment spreads in lists and nested lists of unions), but beyond that it should be more or less complete.

TOTBWF:
Closing this, as it has been subsumed by the subscription support

PR #119, opened 2017-05-31
Add subscription and defer support

TOTBWF:
This PR builds off of #117 to add streaming subscriptions. To create a subscription, you have to do 2 things:

1. Add the subscription to the schema:
```f#let Subscription = 
    Define.SubscriptionObject<Root>(
        name = "Subscription",
        fields = [
            Define.SubscriptionField(
                "name",
                RootType,
                InputType,
                "description",
                [ Define.Input("id", String) ],
                (fun ctx r p -> \*Filter logic here*\ ...))])
```
2. Pass an implementation of `ISubscriptionProvider` to the SchemaConfig 
    ISubscriptionProvider (No need for unsubscribe, that should be handled by disposing of the IObservable):
```f#
type ISubscriptionProvider =
    interface
        /// Registers a new subscription type
        /// Called at schema compilation time
        abstract member Register: Subscription -> unit
        /// Creates an active subscription, and returns the IObservable stream of POCO objects that will be projected on
        abstract member Add : ResolveFieldContext -> obj -> string -> IObservable<obj>
        /// Publishes an event to the subscription system given the identifier of the subscription type
        abstract member Publish<'T> : string -> 'T -> unit
```
A default implementation is provided.


TOTBWF:
Just an update, I fixed a few issues with Spec compliance and null/error propagation and added paths to error messages, as per the spec

Horusiath:
@TOTBWF What is the status on this one?

TOTBWF:
I'm just fixing some edge cases with lists of unions/interfaces and adding some tests, it should be ready within the next few days

TOTBWF:
This should be good to go!

Horusiath:
@johnberzy-bazinga we need to fix CI. It seems to be broken.

johnberzy-bazinga:
Yeah, we're still waiting on travis-ci/travis-build#1060. I'm not sure if there is a work around. I'll take a look at it a bit later.

TOTBWF:
Just for reference, I did some benchmarking with the fixed benchmark code comparing the old and new execution models on both Mono and .NetCore, Here are the results: 
benchmarks
Keep in mind that these are zero-latency fetches.

Horusiath:
Are we ready to merge?

TOTBWF:
Yeah, we are good to go

PR #144, opened 2018-02-23
Adjusting code to remove build warnings

ivelten:
With this Pull Request, I am formatting the code appropriately to remove build warnings.


tests/FSharp.Data.GraphQL.Benchmarks/FSharp.Data.GraphQL.Benchmarks.fsproj
```diff
@@ -4,6 +4,7 @@
     <!-- The benchmarks we use only work on net46 -->
     <TargetFrameworks>net461;netcoreapp1.1</TargetFrameworks>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
ivelten, line 7: The multiple target framework option with paket causes a warning saying that we are referencing 'Microsoft.NETCore.App' explicitly, when this is not necessary. One of the suggested workarounds for that is to disable implicit framework references. This should not be a problem as long as we keep using both frameworks alongside paket.

PR #149, opened 2018-03-05
Adding optional static parameter for introspection schema

ivelten:
With this Pull Request, I am adding an optional static parameter to the client type provider. This parameter can be used to provide the introspection schema JSON, so the type provider is able to work even without being connected to the server.

- [x] Create optional static parameter
- [x] Update sample projects to show how to use it

### DIFF
--- a/samples/client-provider/query.fsx
+++ b/samples/client-provider/query.fsx
@@ -1,4 +1,4 @@
-﻿#r "../../bin/FSharp.Data.GraphQL.Client/FSharp.Data.GraphQL.Client.dll"
+﻿#r "../../bin/FSharp.Data.GraphQL.Client/netstandard2.0/FSharp.Data.GraphQL.Client.dll"
 
 open FSharp.Data.GraphQL
 open System.Collections.Generic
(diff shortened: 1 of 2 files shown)


samples/client-provider/query.fsx
```diff
@@ -8,7 +8,1416 @@ open System.Collections.Generic
 Fable.Core.JsInterop.importAll<unit> "isomorphic-fetch"
 #endif
 
-type MyClient = GraphQLProvider<"http://localhost:8083">
+// Introspection response from the server
+// This can be optionally used as a second parameter to the Provider
+// If using this as a second parameter, connection to the server is not needed to
+// compile the schema and have IntelliSense features when developing
+[<Literal>]
+let introspection = """{
```
ivelten, line 16: I think maybe we should use another type provider to load the introspection from a text file. This avoids having this enormous introspection schema as a string in the code.

Horusiath: Maybe not another type provider, but it can be an optional parmeter of an existing one.

ivelten: All right, I'll be working on something for that.

johnberzy-bazinga:
Combined with #150

PR #150, opened 2018-03-12
[WIP] Creating support for @stream directives

ivelten:
With this Pull Request, I am implementing stream directive support for the server component, as described [here](https://dev-blog.apollodata.com/new-features-in-graphql-batch-defer-stream-live-and-subscribe-7585d0c28b07).

- [x] Mapping streamed fields in the executor
- [x] Create the logic for the stream execution
- [x] Unify streamed and deferred results into a single union case
- [x] Check recursive queries (deferred results inside a streamed result, for example)
- [x] Make tests

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -535,42 +535,6 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
                 | _ ,_ -> raise <| GraphQLException("Path terminated unexpectedly!")
             return res
         }
-    // let deferredResults =
-    //     if ctx.ExecutionPlan.DeferredFields.Length = 0
-    //     then None
-    //     else
-    //         resultTrees
-    //         |> Array.map(AsyncVal.bind(fun tree ->
-    //             ctx.ExecutionPlan.DeferredFields
-    //             |> List.filter (fun d -> (List.head d.Path) = tree.Name)
-    //             |> List.toArray
-    //             |> Array.map(fun d ->
-    //                 let fdef = d.Info.Definition
-    //                 let args = getArgumentValues fdef.Args d.Info.Ast.Arguments ctx.Variables
-    //                 let fieldCtx = { 
-    //                     ExecutionInfo = d.Info
-    //                     Context = ctx
-    //                     ReturnType = fdef.TypeDef
-    //                     ParentType = objdef
-    //                     Schema = ctx.Schema
-    //                     Args = args
-    //                     Variables = ctx.Variables
-    //                 }
-    //                 traversePath d fieldCtx d.Path (AsyncVal.wrap tree) [(List.head d.Path)]
-    //                 |> AsyncVal.bind(Array.map(fun (tree, path) ->
-    //                     asyncVal {
-    //                         // TODO: what to do with deferred errors?
-    //                         let d, _ = treeToDict tree
-    //                         return NameValueLookup.ofList["data", d.Value; "path", upcast path]
-    //                     }) >> AsyncVal.collectParallel))
-    //             |> AsyncVal.collectParallel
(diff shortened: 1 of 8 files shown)


src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -507,86 +494,109 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
...
+                        let! res = buildResolverTree d.Info.ReturnDef fieldCtx fieldExecuteMap t.Value
+                        return! async { return [|res, List.rev(p::pathAcc)|] }
+                    }
+                | [p;"__index"], t ->
+                    asyncVal { 
+                        let! res = buildResolverTree d.Info.ReturnDef fieldCtx fieldExecuteMap t.Value
+                        return! async { return [|res, List.rev(p::pathAcc)|] }
+                    }
+                | [_;"__index";_;"__index"], _ ->
+                    raise <| GraphQLException("Nested defer directives are not supported yet.")
```
ivelten, line 515: We still need to treat this case, which is being handled incorrectly. For now I made a protection to tell the user that nested defer directives inside stream ones are not supported yet.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -507,86 +494,109 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
...
+    let rec traversePath (d : DeferredExecutionInfo) (fieldCtx : ResolveFieldContext) (path: string list) (tree: AsyncVal<ResolverTree>) (pathAcc: string list): AsyncVal<(ResolverTree * string list) []> =
+        asyncVal {
+            let! tree' = tree
+            let! res = 
+                match List.tail path, tree' with
+                | [], t -> 
+                    asyncVal { 
+                        let! res = buildResolverTree d.Info.ReturnDef fieldCtx fieldExecuteMap t.Value
+                        match d.Info.Kind with
+                        | SelectFields [f] -> return! async { return [|res, List.rev (f.Identifier :: pathAcc)|] }
```
ivelten, line 506: I made this treatment to correctly mount the path for deferred fields.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -507,86 +494,109 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
...
+                        | _ -> return! async { return [|res, List.rev ((List.head path)::pathAcc)|] }
+                    }
+                | [p], t ->
```
johnberzy-bazinga: I think this case and the case below can be combined. i.e. ```| ([p;"__index"] | [p]), t -> ...```

ivelten: Yes, it's possible. I will be commiting this change.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -500,36 +500,36 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
             let! res = 
                 match List.tail path, tree' with
                 | [], t -> 
-                    // printfn "%A\n\n%A\n\n%A" d fieldCtx t
-                    // printfn "Done"
+                    printfn "%A\n\n%A\n\n%A" d fieldCtx t
```
Horusiath: You've left print statements uncommented and merged this PR anyway...

johnberzy-bazinga: Oops - I made note of that but forgot to ask @ivelten to make those changes. Good catch.

johnberzy-bazinga: Actually - I think this was removed.

https://github.com/fsprojects/FSharp.Data.GraphQL/blob/dev/src/FSharp.Data.GraphQL.Server/Execution.fs#L497

PR #152, opened 2018-03-26
Stream and defer fixes, removal of unnecessary compiler directives

ivelten:
With this PR, I am implementing several fixes for deferred and stream executions, as well as other minor fixes.

- [x] Fix errors when deferring fields of Union types (exclusive to one type)
- [x] Fix errors when deferring fields of Union types (common to both types)
- [x] Fix errors when deferring fields of Union type lists
- [x] Fix errors when deferring nested defer/stream directives (made a maximum degree of two)
- [x] Fix errors when deferring Enum type lists
- [x] Fix empty result when root object does not have id field specified
- [x] Fix duplicated results when deferring parallel fields (not nested)
- [x] Update and improve tests for all the fixes
- [x] Remove compiler directives regarding reflection calls (they were unified in netcore 2.0)
- [x] Add error handling for errors in deferred results


samples/client-provider/query.fsx
```diff
@@ -1,14 +1,27 @@
...
+// It needs a connection to the server to work in design time
+//type MyClient = GraphQLProvider<url>
+
+// Using the provider like this makes you load the introspection schema from a string
+// The introspection literal is in the literals.fsx file
+//type MyClient = GraphQLProvider<url, introspection>
+
+// Using the provider like this makes you load the introspection schema from a file
+// The file can be load as a relative path to the path of the project or script
+type MyClient = GraphQLProvider<url, "introspection.json">
```
johnberzy-bazinga: If we're going to enable providing the schema statically, I think we should allow for an optional url to be provided at runtime and make the url static parameter optional. Perhaps something similar to what other TPs do?

```
type Schema = GraphQLProvider<"introspection.json">
let myClient = Schema.GetContext(runtimeUrl)
```

What do you think?

Thanks,
John

ivelten: Yes, I agree. I think something like this can be done - I saw it on SqlProvider for connection strings, if I remember well:

```
let [<Literal>] staticUrl = "http://designtimeschema.com"
let runtimeUrl = getConfiguration "serviceUrl"
type provider = GraphQLProvider<staticUrl, "introspection.json">(runtimeUrl)
```

johnberzy-bazinga (approved):

PR #155, opened 2018-04-12
Features: Field Metadata, Extensibility through Execution phase Middleware

ivelten:
With this Pull Request, I am implementing a proposal for extensibility options to the server library. Query executor can now receive middlewares to customize the execution process through many forms, like logging, filtering dangerous queries, change the execution plan, and prevent attacks based on the schema structure.

The middleware is divided in three sub middlewares, representing each phase of the execution: Schema compiling, operation planning and operation execution. A middleware can contain one to three of those sub middlewares.

Also, planning and execution phases can now receive a metadata dictionary to use in customizations. This metadata can be transformed in each middleware, or used to query custom information before passing to the next middleware. On the execution phase, this metadata is sent back through the GQLResponse.

- [x] Implement metadata field for Field Definitions
- [x] Implement middleware interface for Schema Compiling Phase
- [x] Implement middleware interface for Operation Planning Phase
- [x] Implement middleware interface for Execution Phase
- [x] Implement executor middleware interface
- [x] Unit tests for operation execution middleware
- [x] Unit tests for schema compile middleware
- [x] Unit tests for operation planning middleware
- [x] Create middleware for wheightened fields
- [x] Unit tests for weighened fields middleware
- [x] Add middleware interface information to README
- [x] Add weightened fields middleware information to README


src/FSharp.Data.GraphQL.Samples.GiraffeServer/Schema.fs
```diff
@@ -142,14 +142,14 @@ module Schema =
             isTypeOf = (fun o -> o :? Human),
             fieldsFn = fun () -> 
             [
-                Define.Field("id", String, "The id of the human.", fun _ h -> h.Id)
-                Define.Field("name", Nullable String, "The name of the human.", fun _ h -> h.Name)
+                Define.Field("id", String, "The id of the human.", fun _ (h : Human) -> h.Id)
```
ivelten: For some reason, implementing the possibility of weighted fields as an argument to the Define.Field makes some of those functions to not recognize the correct argument type. I need to work on that somehow.

src/FSharp.Data.GraphQL.Server/Executor.fs
```diff
@@ -1,16 +1,22 @@
...
+    ExecutionPlan * 'Root option * Map<string, obj> option -> Async<GQLResponse>
 
-type Executor<'Root> (schema: ISchema<'Root>) = 
+type ExecutionMiddlewareFunc<'Root> = 
+    ExecutionPlan * 'Root option * Map<string, obj> option -> ExecutionFunc<'Root> -> Async<GQLResponse>
+
+type IExecutionMiddleware<'Root> =
+    abstract member AsyncResolve : ExecutionMiddlewareFunc<'Root>
+
+type Executor<'Root>(schema: ISchema<'Root>, middlewares : IExecutionMiddleware<'Root> seq) =
```
ivelten: One of the ways to inject middleware is by implementing IExecutionMiddleware<'Root> interfaces for each middleware needed.

The interface does have one single function, AsyncResolve, which receives the arguments of a execution process (Execution plan, root type definition, data and variables) and the next execution call (if the middleware needs to pass on to the next middleware).

src/FSharp.Data.GraphQL.Server/Executor.fs
```diff
@@ -59,6 +65,44 @@ type Executor<'Root> (schema: ISchema<'Root>) =
...
+
+    let executeWithMiddlewares (executionPlan: ExecutionPlan, data: 'Root option, variables: Map<string, obj> option) =
+        middlewares
+        |> Seq.map (fun m -> m.AsyncResolve)
+        |> List.ofSeq
+        |> runMiddlewares executionPlan data variables
+
+    new(schema) = Executor(schema, middlewares = Seq.empty)
+
+    new(schema, middlewareFuncs : ExecutionMiddlewareFunc<'Root> seq) =
```
ivelten: If the user wants to use plain functions as middleware, they can through the use of this constructor. Works exactly as the interface option, except that we just use the AsyncResolve function, which is this function.

src/FSharp.Data.GraphQL.Server/Executor.fs
```diff
@@ -59,6 +65,44 @@ type Executor<'Root> (schema: ISchema<'Root>) =
...
+        | [x] -> x (executionPlan, data, variables) (fun (plan, data, variables) -> runMiddlewares plan data variables [])
+        | x :: xs -> x (executionPlan, data, variables) (fun (plan, data, variables) -> runMiddlewares plan data variables xs)
+
+    let executeWithMiddlewares (executionPlan: ExecutionPlan, data: 'Root option, variables: Map<string, obj> option) =
+        middlewares
+        |> Seq.map (fun m -> m.AsyncResolve)
+        |> List.ofSeq
+        |> runMiddlewares executionPlan data variables
+
+    new(schema) = Executor(schema, middlewares = Seq.empty)
```
ivelten: The default constructor is maintained for when no middleware is wanted.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -1536,6 +1543,17 @@ and DirectiveDef =
       /// Array of arguments defined within that directive.
       Args : InputFieldDef [] }
 
+/// Field metadata definition.
+and Metadata() =
```
ivelten: This is the metadata implementation, just a facade for a Dictionary<string, obj>.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -2306,13 +2325,17 @@ module SchemaDefinitions =
         /// </summary>
         /// <param name="name">Field name. Must be unique in scope of the defining object.</param>
         /// <param name="typedef">GraphQL type definition of the current field's type.</param>
-        static member Field(name : string, typedef : #OutputDef<'Res>) : FieldDef<'Val> = 
+        static member Field(name : string, typedef : #OutputDef<'Res>, ?weight : float) : FieldDef<'Val> =
```
ivelten: Fields now can be weighted. This creates a "weight" value inside the metadata.

tests/FSharp.Data.GraphQL.Tests/QueryWeightMiddlewareTests.fs
```diff
@@ -0,0 +1,365 @@
...
+            name = "Query",
+            fields = 
+                [ Define.Field("A", Nullable AType, "A Field", [ Define.Input("id", Int) ], resolve = fun ctx _ -> getA (ctx.Arg("id")))
+                  Define.Field("B", Nullable BType, "B Field", [ Define.Input("id", Int) ], resolve = fun ctx _ -> getB (ctx.Arg("id"))) ])
+    let schema = Schema(Query)
+    let middlewares = [ QueryWeightMiddleware() :> IExecutorMiddleware ]
+    Executor(schema, middlewares)
+
+let execute (query : Document) =
+    executor.AsyncExecute(query, meta = Metadata.QueryWeightThreshold(2.0)) |> sync
```
ivelten: By making the query threshold have a weight of 2, it means that each one of the cases of the SubjectType have a weight of 1 each. If we make a query on "subject" field of A or B, without using fragments (for example, extract common fields to both cases), the total weight for this field will be 2, meaning that it can't be queried again in a nested way.

tests/FSharp.Data.GraphQL.Tests/QueryWeightMiddlewareTests.fs
```diff
@@ -0,0 +1,365 @@
...
+    let result = execute query
+    match result with
+    | Direct (data, errors) ->
+        empty errors
+        data.["data"] |> equals (upcast expected)
+    | _ -> fail "Expected Direct GQLResponse"
+    result.Metadata.TryFind<float>("queryWeightThreshold") |> equals (Some 2.0)
+
+[<Fact>]
+let ``Inline fragment query : Should not pass when above threshold``() =
```
ivelten: If we use fragments though, and specify the field in only one of them, then we have a weight of 1 for it. In this case, the query still fails to execute because intentionally we queried a nested subjects on this fragment.

src/FSharp.Data.GraphQL.Server/Planning.fs
```diff
@@ -310,7 +305,7 @@ let private planVariables (schema: ISchema) (operation: OperationDefinition) =
                 { VarDef.Name = vname; TypeDef = idef; DefaultValue = vdef.DefaultValue }
             | _ -> raise (MalformedQueryException (sprintf "GraphQL query defined variable '$%s' of type '%s' which is not an input type definition" vname (tdef.ToString()))))
 
-let internal planOperation documentId (ctx: PlanningContext) (operation: OperationDefinition) : ExecutionPlan =
+let internal planOperation (ctx: PlanningContext) : ExecutionPlan =
```
ivelten, line 308: Here I refactored  PlanningContext to hold documentId and the operation definition as well. This makes this object be able to hold all needed information to plan an operation. Also, makes the middleware signature simpler.

src/FSharp.Data.GraphQL.Server.Middlewares/TypeSystemExtensions.fs
```diff
@@ -0,0 +1,16 @@
+namespace FSharp.Data.GraphQL.Server.Middlewares
+
+open FSharp.Data.GraphQL.Types
+
+[<AutoOpen>]
+module TypeSystemExtensions =
```
ivelten: This module is meant to add extensions to add metadata used by the QueryWeightMiddleware.
Maybe we should change the namespace to be FSharp.Data.GraphQL.Server, so we have access to those extensions without manually needing to add this namespace to the calling code.

src/FSharp.Data.GraphQL.Samples.GiraffeServer/HttpHandlers.fs
```diff
@@ -64,15 +64,15 @@ module HttpHandlers =
             printfn "Received query: %s" query
             printfn "Received variables: %A" variables
             let query = query |> removeSpacesAndNewLines
-            let result = Schema.executor.AsyncExecute(query, variables = variables, data = Schema.root) |> Async.RunSynchronously
+            let result = Schema.executor.AsyncExecute(query, variables = variables, data = Schema.root, meta = Schema.executionMetadata) |> Async.RunSynchronously
```
ivelten: Metadata can now be passed on each execution call. In this example, it is fixed, but now it could add per-request information, since this method is being called on an endpoint call.

src/FSharp.Data.GraphQL.Samples.GiraffeServer/Schema.fs
```diff
@@ -140,40 +142,37 @@ module Schema =
...
+            fieldsFn = fun () ->
             [
                 Define.Field("id", String, "The id of the human.", fun _ h -> h.Id)
                 Define.Field("name", Nullable String, "The name of the human.", fun _ h -> h.Name)
                 Define.Field("friends", ListOf (Nullable CharacterType), "The friends of the human, or an empty list if they have none.",
-                    fun _ h -> 
-                        h.Friends
-                        |> List.map getCharacter 
-                        |> List.toSeq)
+                    fun _ (h : Human) -> h.Friends |> List.map getCharacter |> List.toSeq).WithQueryWeight(0.5)
```
ivelten, line 149: I added the query weight threshold middleware in the giraffe sample project. Friends now have a weight of 0.5 each, and the maximum weight is 2.0.

src/FSharp.Data.GraphQL.Server.Middlewares/AssemblyInfo.fs
```diff
@@ -0,0 +1,17 @@
+﻿// Auto-Generated by FAKE; do not edit
+namespace System
+open System.Reflection
+
+[<assembly: AssemblyTitleAttribute("FSharp.Data.GraphQL.Server.Middlewares")>]
```
ivelten, line 5: Not sure about this file. Should we have an assembly info in each project even though it is generated by the build?

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -637,7 +642,14 @@ and DeferredExecutionInfoKind =
     | DeferredExecution
     | StreamedExecution
 
-and ExecutionPlan = 
+/// The context used to hold all the information for a schema compiling proccess.
+and SchemaCompileContext =
```
ivelten, line 703: This context is new and will be used in the compile schema phase.

tests/FSharp.Data.GraphQL.Tests/OperationExecutionMiddlewareTests.fs
```diff
@@ -0,0 +1,86 @@
...
+
+let ast = parse """{
+        testData {
+            a
+            b
+            c
+        }
+    }"""
+
+let operationMiddleware (ctx : ExecutionContext) (next : ExecutionContext -> AsyncVal<GQLResponse>) =
```
ivelten: To test the middleware interface, I made a simple middleware that takes out one of the fields of the schema on the execution plan ("c" field). After making the needed changes on the execution plan, it calls the next middleware to continue the execution.

src/FSharp.Data.GraphQL.Server.Middlewares/DefineExtensions.fs
```diff
@@ -0,0 +1,29 @@
...
+
+open FSharp.Data.GraphQL
+open FSharp.Data.GraphQL.Types
+
+/// Contains extensions for the Define module.
+[<AutoOpen>]
+module DefineExtensions =
+    type Define with
+        /// <summary>
+        /// Creates a query weight middleware, with the specified query weight threshold.
```
Horusiath, line 11: The comment doesn't describe the default value of the threshold.

src/FSharp.Data.GraphQL.Server.Middlewares/ObjectListFilter.fs
```diff
@@ -0,0 +1,52 @@
...
+    | GreaterThan of FieldFilter<System.IComparable>
+    | LessThan of FieldFilter<System.IComparable>
+    | StartsWith of FieldFilter<string>
+    | EndsWith of FieldFilter<string>
+    | Contains of FieldFilter<string>
+    | Field of FieldFilter<ObjectListFilter>
+
+/// Contains operators for building and comparing ObjectListFilter values.
+[<AutoOpen>]
+module ObjectListFilterOperators =
```
Horusiath: Don't over-introduce custom operators. They're leveraging the learning curve and code readability sky-high. Maybe it would be just easier to convert those values directly from F# quotation expressions?

Also for custom operators a good idea is to contain them in separate module like `Operators` that is not auto-opened. This way user can decide, whether to use them or not.

src/FSharp.Data.GraphQL.Server.Middlewares/DefineExtensions.fs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        /// <param name="threshold">A float value representing the maximum threshold for any query analyzed by the middleware.</param>
+        static member QueryWeightMiddleware(?threshold : float) : IExecutorMiddleware =
+            match threshold with
```
Horusiath: You should be able to fall through this value to `QueryWeightMiddleware` constructor.

src/FSharp.Data.GraphQL.Server.Middlewares/QueryWeightMiddleware.fs
```diff
@@ -0,0 +1,57 @@
...
+                        let fields = typeFields |> Map.toList |> List.collect (fun (_, v) -> v)
+                        let (pass, current) = checkThreshold current fields
+                        if pass then checkThreshold current xs else (false, current)
+                     | _ -> 
+                        checkThreshold current xs
+        checkThreshold 0.0 fields
+    let middleware (ctx : ExecutionContext) (next : ExecutionContext -> AsyncVal<GQLResponse>) =
+            match threshold with
+            | Some t -> 
+                if ctx.Metadata.ContainsKey(MetadataKeys.QueryWeightMiddleware.QueryWeightThreshold)
```
Horusiath: You could use F# map and simply call `Map.add` instead - it has upsert semantics.

src/FSharp.Data.GraphQL.Server.Middlewares/TypeSystemExtensions.fs
```diff
@@ -0,0 +1,158 @@
...
+        /// <param name="middleware">The middleware function for the field resolve.</param>
+        member this.WithResolveMiddleware<'Res>(middleware : FieldResolveMiddleware<'Val, 'Res>) : FieldDef<'Val> =
+            upcast CustomResolveFieldDefinition(this, middleware)
+
+        /// <summary>
+        /// Adds metadata information to existing field definition, containing
+        /// the query weight value for it. This value is used by the QueryWeightMiddleware to calculate a query weight.
+        /// </summary>
+        /// <param name="weight">A float value representing the weight that this field have on the query.</param>
+        member this.WithQueryWeight(weight : float) =
```
Horusiath: If I remember correctly, original `FieldDef<>` is immutable and behaves like that in all methods. This method mutates the object instead of returning a new copy of it. Please use F# maps.

src/FSharp.Data.GraphQL.Server.Middlewares/TypeSystemExtensions.fs
```diff
@@ -0,0 +1,158 @@
...
+    type Metadata with
+        /// <summary>
+        /// Adds metadata information to the current metadata definition, containing
+        /// the maximum query weight (threshold) for query execution.
+        /// This value is used by the QueryWeightMiddleware to measure if a query weight is 
+        /// below a defined threshold.
+        /// </summary>
+        /// <param name="threshold">A float value, representing the threshold weight.</param>
+        member this.WithQueryWeightThreshold(threshold : float) =
+            this.Add(MetadataKeys.QueryWeightMiddleware.QueryWeightThreshold, threshold); this
```
Horusiath: Same as with `FieldDef<>`. Try to keep objects immutable.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -105,37 +120,41 @@ type NameValueLookup(keyValues: KeyValuePair<string, obj> []) =
     interface IEquatable<NameValueLookup> with
         member x.Equals(other) = structEq x other
     interface System.Collections.IEnumerable with
-        member x.GetEnumerator() = (kvals :> System.Collections.IEnumerable).GetEnumerator()
+        member __.GetEnumerator() = (kvals :> System.Collections.IEnumerable).GetEnumerator()
```
Horusiath, line 123: 👍

src/FSharp.Data.GraphQL.Server/Executor.fs
```diff
@@ -1,142 +1,232 @@
...
+    /// Contains the function that holds the schema compile phase for this middleware.
+    abstract member CompileSchema : SchemaCompileMiddleware option
+    /// Contains the function that holds the operation planning phase for this middleware.
+    abstract member PlanOperation : OperationPlanningMiddleware option
+    /// Contains the function that holds the operation execution phase for this middleware.
+    abstract member ExecuteOperationAsync : OperationExecutionMiddleware option
+
+/// A simple implementation for an executor middleware interface.
+/// It can hold one to three functions for each sub phase of the execution process.
+type ExecutorMiddleware(?compile, ?plan, ?execute) =
```
Horusiath: I think, this could be actually a discriminated union with 3 cases, 1 for every phase.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
Horusiath: Try to keep it immutable as it's being used within immutable structures. Mixing mutable and immutable types is recipe for a headache ;) Especially as you're using `Empty` property.

Horusiath (changes requested):
Please make some documentation. Feature not documented doesn't exists. This one is not described on github, neither in docs. It makes it hard to review.

Some notes:

- In general I think, that this pull request could be split in two: middleware + filters. 
- The general idea behind middleware for 3 phases (compilation, planing, execution) seems to be right. I've got some idea, that we could make use of [TypeShape](https://github.com/eiriktsarpalis/TypeShape) visitor pattern here. But it's not blocking.
- Try to keep type definitions immutable. Especially when they are composed within other immutable types.
- Don't introduce new operators, as they are making things harder. When you feel you need to, think about keeping them internal.
- 👍 for some nice code cleaning ;)

src/FSharp.Data.GraphQL.Server.Middlewares/DefineExtensions.fs
ivelten: Fixed that. I commented about what happens when no value is provided.

src/FSharp.Data.GraphQL.Server.Middlewares/ObjectListFilter.fs
ivelten: I did the second suggestion of yours. I placed them inside ObjectListFilter.Operators module, and it is not auto opened.

src/FSharp.Data.GraphQL.Server.Middlewares/QueryWeightMiddleware.fs
ivelten: I changed all metadata objects to immutable ones as suggested, and now new copies are being generated each time they need to be modified.

src/FSharp.Data.GraphQL.Server.Middlewares/TypeSystemExtensions.fs
ivelten: Done.

src/FSharp.Data.GraphQL.Server/Executor.fs
ivelten: I did that change. I am not sure though, if the Interface definition is useful. For now I am only using functions itselves as middlewares.

Horusiath (approved):

PR #168, opened 2018-07-02
Input definition being nullable if default value is provided

ivelten:
This PR is a fix proposal for the issue reported [here](https://github.com/fsprojects/FSharp.Data.GraphQL/issues/167).

It makes a input field definition automatically as a nullable value, when a default value is provided.


src/FSharp.Data.GraphQL.Server/Schema.fs
```diff
@@ -107,10 +107,17 @@ type Schema<'Root> (query: ObjectDef<'Root>, ?mutation: ObjectDef<'Root>, ?subsc
...
-        { Name = inputDef.Name
-          Description = inputDef.Description
-          Type = introspectTypeRef false namedTypes inputDef.TypeDef
-          DefaultValue = inputDef.DefaultValue |> Option.map string }
+        match inputDef.DefaultValue with
+        | Some defaultValue ->
+            { Name = inputDef.Name
+              Description = inputDef.Description
+              Type = introspectTypeRef true namedTypes inputDef.TypeDef
+              DefaultValue = Some defaultValue |> Option.map string }
```
Horusiath: Use `Some (string defaultValue)` - Option.map can potentially create another option, while there's no need for that.

src/FSharp.Data.GraphQL.Server/Schema.fs
```diff
@@ -107,10 +107,17 @@ type Schema<'Root> (query: ObjectDef<'Root>, ?mutation: ObjectDef<'Root>, ?subsc
...
     let introspectInput (namedTypes: Map<string, IntrospectionTypeRef>) (inputDef: InputFieldDef) : IntrospectionInputVal =
-        { Name = inputDef.Name
-          Description = inputDef.Description
-          Type = introspectTypeRef false namedTypes inputDef.TypeDef
-          DefaultValue = inputDef.DefaultValue |> Option.map string }
+        match inputDef.DefaultValue with
+        | Some defaultValue ->
+            { Name = inputDef.Name
+              Description = inputDef.Description
+              Type = introspectTypeRef true namedTypes inputDef.TypeDef
```
Horusiath: This can actually be written as `(Option.isSome inputDef.DefaultValue)` - in that case you won't need pattern match at all.

Horusiath (changes requested):

Horusiath:
2 more issues:

1. You're committing 10 000 LOC for the sake of having paket lock file - this seems unjustified (especially since the change itself is 1 line of code).
2. There are not tests to actually verify that this change solved an issue.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -228,19 +222,6 @@ let private createFieldContext objdef argDefs ctx (info: ExecutionInfo) =
       Args = args
       Variables = ctx.Variables }         
                 
-
-let private optionCast (value: obj) =
```
ivelten: I took out those functions and put them on the Shared library, because I wanted to use them on another piece of code for this issue.

ivelten:
@Horusiath could you give another look into this PR? I have made all the changes you asked for.

Horusiath (approved):

PR #169, opened 2018-07-04
Introspection test fixes

ivelten:
With this PR, I am fixing some skipped (and broken tests).
Also, fixing an error on the structural equality of `NameValueLookup`.

- [x] Fix test 'Introspection schema should be serializable back and forth using json'
- [x] Fix test 'Introspection executes an introspection query'
- [x] Fix `NameValueLookup` wrong structural equality


src/FSharp.Data.GraphQL.Shared/Introspection.fs
```diff
@@ -166,7 +166,7 @@ let rec __Type =
             | Some name ->
                 let found = findIntrospected ctx name
                 found.PossibleTypes |> Option.map Array.toSeq)
-        Define.Field("enumValues", Nullable (ListOf __EnumValue), description = null,
```
ivelten, line 169: This was causing the introspection schema to have `Some null` as description instead of `None`, so I created an overload that does not ask for a description instead.

src/FSharp.Data.GraphQL.Samples.GiraffeServer/Schema.fs
```diff
@@ -158,8 +158,8 @@ module Schema =
             isTypeOf = (fun o -> o :? Droid),
             fieldsFn = fun () ->
             [
-                Define.Field("id", String, "The id of the droid.", fun _ d -> d.Id)
-                Define.Field("name", Nullable String, "The name of the Droid.", fun _ d -> d.Name)
+                Define.Field("id", String, "The id of the droid.", fun _ (d : Droid) -> d.Id)
```
ivelten, line 164: Unfortunately, creating an overload of `Define.Field` that asks no description, but arguments made this change necessary for it to compile.

tests/FSharp.Data.GraphQL.Tests/IntrospectionTests.fs
```diff
@@ -21,15 +20,119 @@ type IntrospectionData = {
...
-
-[<Fact(Skip="FIXME: investigate reason of failure")>]
+[<Fact>]
 let ``Introspection schema should be serializable back and forth using json`` () =
     let root = Define.Object("Query", [ Define.Field("onlyField", String) ])
     let schema = Schema(root)
-    let introResult = Executor(schema).AsyncExecute(Introspection.introspectionQuery) |> sync
-    let json = toJson introResult
-    let deserialized:IntrospectionData = Helpers.fromJson json
-    deserialized.Data.__schema |> equals (schema :> ISchema).Introspected
```
ivelten, line 32: This test was not working mainly because of two factors:

1. The introspection query inside `Introspection.introspectionQuery` does not ask for all fields that are serialized.
2. The `equals` comparison was being made with the `GQLResponse` itself, which is not serialized at all.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -69,15 +69,17 @@ type NameValueLookup(keyValues: KeyValuePair<string, obj> []) =
...
             x.Buffer
             |> Array.forall2 (fun (a: KeyValuePair<string, obj>) (b: KeyValuePair<string, obj>) ->
                 if a.Key <> b.Key then false
                 else 
                     match a.Value, b.Value with
-                    | :? NameValueLookup, :? NameValueLookup as o -> structEq (downcast fst o) (downcast snd o)
-                    | :? seq<obj>, :? seq<obj> -> Seq.forall2 (=) (a.Value :?> seq<obj>) (b.Value :?> seq<obj>)
+                    | (:? NameValueLookup as x), (:? NameValueLookup as y) -> structEq x y
+                    | (:? seq<obj> as x), (:? seq<obj> as y) -> 
+                        if Seq.length x <> Seq.length y then false else Seq.forall2 (=) x y
```
ivelten, line 76: `NameValueLookup` structural equality was wrong when this assertion was made, ignoring differences in sequence lengths. If the first sequence (x) was a part of the second (contains a set of the first items), the test considered it equal, even if there was missing items in the end.

PR #172, opened 2018-08-09
Changing context input for subscriptions

ivelten:
With this PR, I am changing the `SubscriptionFieldDefinition` object to be able to have a generic typed output, making it not required to be a GraphQL type. This will be a breaking change on the filter function and the provider implementations as well, since the filter function now returns the output as an option instead of a boolean value.

Also, implementations for asynchronous subscriptions fields were made.

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -712,7 +712,6 @@ let private executeSubscription (resultSet: (string * ExecutionInfo) []) (ctx: E
             match err with
             | [] -> NameValueLookup.ofList["data", data.Value] :> Output
             | _ -> NameValueLookup.ofList["data", data.Value; "errors", upcast (formatErrors err)] :> Output)
-        |> AsyncVal.map(fun d -> printfn "Async dict %A" d;d)
         |> AsyncVal.toAsync
         |> Observable.ofAsync)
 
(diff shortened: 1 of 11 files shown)


johnberzy-bazinga (commented):
Can we also update the samples with the applicable subscription changes.

Thanks

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -84,14 +84,15 @@ type NameValueLookup(keyValues: KeyValuePair<string, obj> []) =
...
         | :? NameValueLookup as lookup ->
-            sb.Append("{ ") |> ignore
-            lookup.Buffer
-            |> Array.iter (fun kv -> 
-                sb.Append(kv.Key).Append(": ") |> ignore
-                stringify sb (deep+1) kv.Value
-                sb.Append(",\r\n") |> ignore
-                pad sb deep)
-            sb.Remove(sb.Length - 4 - deep, 4 + deep).Append(" }") |> ignore
+            if lookup.Count > 0 then
```
ivelten, line 87: `stringify` was generating an error when empty NameValueLookups where used in the `o` parameter. So I just added a conditional to skip it when no value is present.

johnberzy-bazinga:
LGTM. We should update the samples in another PR.

johnberzy-bazinga:
Travis issue is unrelated to this PR

PR #175, opened 2018-08-10
Changing subscription to delay async field evaluation to the last moment

ivelten:
With this PR, I am fixing the subscription for asynchronous fields to be evaluated on the last moment.
Also, adding the release notes for 0.0.5-beta.


johnberzy-bazinga:
LGTM. Thanks!

PR #178, opened 2018-09-12
Async versions of Register and Publish of subscription providers

ivelten:
With this PR, I am changing `Register` and `Publish<'T>` methods of `ISubscriptionProvider` and `ILiveFieldSubscriptionProvider` to be asynchronous (also renaming them to have `Async` suffix). To maintain backwards compatibility for the sync methods, they are now extensions that runs their async versions synchronously.


johnberzy-bazinga:
LGTM. thanks!

PR #181, opened 2018-10-10
File upload support

ivelten:
With this PR, I am adding an experimental tool package for working with AspNetCore using this [multipart request spec](https://github.com/jaydenseric/graphql-multipart-request-spec) to enable a file upload feature using GraphQL.

The idea behind the component is to provide a MultipartRequest object that will be parsed from a MultipartReader. This parsing operation will automatically fill the file into the variables object, enabling user code to fetch the stream from the resolver function of a mutation field, for example.

- [x] Create an input object for file upload
- [x] Create a MultipartRequest object
- [x] Create a routine to read multipart requests and parse them to the MultiparseRequest object
- [x] Automated tests
- [x] Update sample server
- [x] Create nuget package
- [x] Create documentation


src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
```diff
@@ -0,0 +1,99 @@
...
+        let operations =
+            match JToken.Parse(operations) with
+            | :? JArray as ops -> ops.ToObject<Operation list>()
+            | :? JObject as op -> [ op.ToObject<Operation>() ]
+            | _ -> failwith "Unexpected operations value."
+        let map = 
+            map 
+            |> Seq.map (fun x -> x.Key, x.Value |> Seq.tryHead)
+            |> Seq.filter (fun (_, v) -> Option.isSome v)
+            |> Seq.map (fun (k, v) -> k, v.Value)
```
Horusiath: Use `Seq.choose` instead:

```fsharp
map
|> Seq.choose (fun x -> x.Value |> Seq.tryHead |> Option.map (fun v -> (k, v))
|> Map.ofSeq
```

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
```diff
@@ -0,0 +1,99 @@
...
+            |> Map.ofSeq
+        let files =
+            files
+            |> Seq.map (fun x -> x.Key, x.Value)
```
Horusiath: You can use `KeyValue` acive pattern: `Seq.map (|KeyValue|)`.

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
```diff
@@ -0,0 +1,99 @@
...
+            let findFile (varName : string) (varValue : obj) =
+                let pickFromMap varName =
+                    map |> Map.tryPick (fun k v -> if v = varName then files |> Map.tryFind k else None)
+                match varValue with
+                | :? seq<obj> as values ->
+                    values
+                    |> Seq.mapi (fun valueIndex value ->
+                        let varName =
+                            match operationIndex with
+                            | Some operationIndex -> sprintf "%i.variables.%s.%i" operationIndex varName valueIndex
```
Horusiath: Keep in mind that sprintf is an expensive operation.

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
```diff
@@ -0,0 +1,99 @@
...
+    /// Reads a GraphQL multipart request from a MultipartReader.
+    let read (reader : MultipartReader) = 
+        let getName (section : MultipartSection) =
+            match section.AsFormDataSection() with
+            | null -> section.AsFileSection().Name
+            | x -> x.Name
+        let readText (section : MultipartSection) =
+            section.AsFormDataSection().GetValueAsync()
+            |> Async.AwaitTask
+            |> Async.RunSynchronously
```
Horusiath: `Async.RunSynchronously`? (┛◉Д◉)┛彡┻━┻

Horusiath (changes requested):
Few minor stylish issues, and few major ones (mostly around the area of blocking run of asyncs and sprintf). 

Also I'm worried about relying on JSON.NET more and more, but as this is a companion library which already has ASP.NET Core dependencies, I think we're eventually ok with that.

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
ivelten: I switched that for `String.Format`, as it seems better by this [comparison](http://www.fssnip.net/mQ).

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
ivelten: OK, true, that was bad. Solved.

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
ivelten: Done.

src/FSharp.Data.GraphQL.Server.Middlewares.AspNetCore/MultipartRequest.fs
ivelten: Done.

ivelten:
@Horusiath I did all the changes you asked for. About JSON.NET dependency, yeah, as this component is designed for ASP.NET Core and Microsoft already adopted this library, I don't think this is an issue in this case.

Rajivhost:
Apparently done, I'm waiting this to be merged :)

xperiandri:
@ivelten, does this functionality work?
Could you rebase it onto the latest dev?

ivelten:
@xperiandri it was working when I finished the PR. If everyone decides to restart this PR, I can try rebasing and/or adjusting minor issues on it next weekend.

xperiandri:
@aggieben, @johnberzy-bazinga, any objections?

aggieben:
I don't object to the PR, but I'd think you'd want to rebase before merging

johnberzy-bazinga:
I'd hold off on rebasing this. I'm working on a middleware that uses `System.Text.Json` that also supports File uploads.  The goal is to remove the boilerplate by supporting something like:

`app.UseGraphQL(...)`

It's close to being done, but on problem I ran into with `System.Text.Json` is it doesn't have anything analogues to `JObject.FromObject` so it's trickier to support the replacements by jsonpath that are needed for multipart file uploads.

xperiandri:
I suppose that according to the sample in this repo it is already done, right?

aggieben:
Don't we want the server-side implementation in this PR?

xperiandri:
It was there as a sample but I broke it when I implemented getting variables from JsonDocument

PR #182, opened 2018-10-11
Adding assembly redirect resolve inside type provider

ivelten:
With this PR, I am adding a routine to treat assembly redirects inside the GraphQL TypeProvider, so new versions of dependencies can be resolved at runtime.


ivelten:
Closing this as those changes don't solve all the problems that are happening with the client provider ATM. I am opening a new pull request for that, redoing the client component using the actual type provider framework.

PR #188, opened 2018-11-13
Defer stream adjusts

ivelten:
With this PR, I am adding a behavior to the server component to help with old clients. When deferring results, the server should also send the deferred fields in the direct result, with an empty value.


johnberzy-bazinga:
LGTM. Thanks

PR #191, opened 2018-11-21
Deferring fields under a root interface field

ivelten:
This Pull Request solves an issue where fields under an interface that is inside the root object are not deferred.

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -539,24 +539,10 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
                     }
                 | [String p], t ->
                     asyncVal {
-                        let resolve value =
-                            asyncVal {
-                                let! res = buildResolverTree d.Info.ReturnDef fieldCtx fieldExecuteMap value
-                                match res with
-                                | ResolverError _ -> return! async { return [||] } // A deferred fragment that was not found, just ignore it
-                                | _ -> return! async { return [|res, List.rev((p :> obj)::pathAcc)|] }
-                            }
-                        let resolveObject n =
-                            asyncVal {
-                                let head = d.Info.Definition.Name
-                                let next = n.Children |> Array.tryFind (fun c' -> c'.Name = head)
-                                match next with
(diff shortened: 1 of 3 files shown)


src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -539,24 +539,10 @@ let private executeQueryOrMutation (resultSet: (string * ExecutionInfo) []) (ctx
                     }
                 | [String p], t ->
                     asyncVal {
-                        let resolve value =
```
ivelten, line 542: This piece of code can be safely removed, since it was trying to handle the wrong path built by the planning function.

src/FSharp.Data.GraphQL.Server/Planning.fs
```diff
@@ -208,7 +208,7 @@ let rec private plan (ctx : PlanningContext) (stage : PlanningStage) : PlanningS
         let inner, deferredFields', path' = plan ctx ({ info with ParentDef = info.ReturnDef; ReturnDef = downcast returnDef; Identifier = "__index" }, deferredFields, "__index"::info.Identifier::path)
         { info with Kind = ResolveCollection inner }, deferredFields', path'
     | Abstract _ -> 
-        planAbstraction ctx info.Ast.SelectionSet (info, deferredFields, path) (ref []) None 
+        planAbstraction ctx info.Ast.SelectionSet (info, deferredFields, info.Identifier::path) (ref []) None
```
ivelten, line 211: This is the root of the problem. Interface planning wasn't adding its identifier to the path, causing unexpected entries in the `traversePath` function.

tests/FSharp.Data.GraphQL.Tests/Relay/NodeTests.fs
```diff
@@ -82,9 +133,9 @@ let ``Node with global ID gets correct record`` () =
     let expected2 =    
       NameValueLookup.ofList [
         "model", upcast "Tesla S" ]
-    execAndValidateNode query1 expected1
+    execAndValidateNode query2 expected2 None
```
ivelten, line 136: This test was broken, validating two times the first case, instead of the two cases.

johnberzy-bazinga:
👍 

CI issue is unrelated

PR #200, opened 2018-12-06
Creating subscription publish model that chooses filter functions

ivelten:
This is a draft Pull Request. The feature is still being discussed.

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
+++ b/src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
@@ -279,22 +279,27 @@ module Introspection =
 
 /// Represents a subscription as described in the schema.
 type Subscription = {
-    /// The name of the subscription type in the schema
-    Name: string
-    /// Filter function, used to determine what events we will propagate
-    /// The 1st obj is the boxed root value, the second is the boxed value of the input object
-    Filter: (ResolveFieldContext -> obj -> obj -> Async<obj option>)
-}
+    /// The name of the subscription type in the schema.
+    Name : string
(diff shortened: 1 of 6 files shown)


src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -1970,17 +1965,21 @@ and TypeMap() =
...
 
+module Tags =
+    let from (x : #Tag) : Tag seq = Seq.singleton (upcast x)
+    let fromSeq x : Tag seq = x |> Seq.map (fun t -> upcast t)
+
 [<AutoOpen>]
 module SubscriptionExtensions =
     type ISubscriptionProvider with
         member this.Register subscription =
             this.RegisterAsync subscription |> Async.RunSynchronously
```
johnberzy-bazinga: I know this will break backwards compat, but I believe we should change all the XyzAsync methods to AsyncXyx while it's still early.

johnberzy-bazinga:
Other than minor name change, this LGTM 👍

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
ivelten: Agreed - changes pushed.

johnberzy-bazinga: Can we also change them for the ILiveSubscriptionProvider? thanks

ivelten: Done.

PR #201, opened 2018-12-08
Defer/Stream parallel issues

ivelten:
A deferred result or streamed result item should be sent as soon as it finishes processing (resolver function applies). This PR means to fix those issues.

- [x] Validation tests
- [x] When having more than one deferred/streamed field, each field should be deferred/streamed as soon as it finishes computing
- [x] When having a streamed field, each list item should be returned as soon as it finishes computing

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -268,7 +268,10 @@ module ResolverTree =
                 leafOp ((leaf.Name :> obj)::path) leaf
             | ResolverError err ->
                 let origin = (err.PathToOrigin |> List.rev)
-                let path' = if err.Name <> "__index" then origin@((err.Name :> obj)::path) else origin@path
+                let toString x = x.ToString()
+                let head = List.tryHead path |> Option.map toString
+                let shouldAdd = err.Name <> "__index" && Some (err.Name) <> head
+                let path' = if shouldAdd then origin@((err.Name :> obj)::path) else origin@path
                 errorOp path' err
             | ResolverObjectNode node ->
                 let path' = if node.Name <> "__index" then (node.Name :> obj)::path else path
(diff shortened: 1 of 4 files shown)


src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -259,23 +259,9 @@ type ResolverTree =
         | ResolverListNode l -> l.Value
 and ResolverLeaf = { Name: string; Value: obj option }
 and ResolverError = { Name: string; Message: string; PathToOrigin: obj list }
-and ResolverNode = { Name: string; Value: obj option; Children: ResolverTree [] }
+and ResolverNode = { Name: string; Value: obj option; Children: AsyncVal<ResolverTree> [] }
 
 module ResolverTree =
-    let fold leafOp errorOp nodeOp listOp =
```
ivelten: We just use pathFold, so I removed this function.

ivelten, line 265: We don't use this function at all, so I removed it.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -286,34 +272,82 @@ module ResolverTree =
...
-                let ts = node.Children |> Array.mapi(fun i c -> helper ((i :> obj)::path') c)
+                let mapper (i : int) (c : AsyncVal<ResolverTree>) = asyncVal {
+                    let! c' = c
+                    return helper ((box i)::path') c'
+                }
+                let ts = node.Children |> Array.mapi mapper
                 listOp path' node.Name node.Value ts
         helper []
 
+let private foldChildren (children : AsyncVal<AsyncVal<KeyValuePair<string, obj> * Error list>> []) =
```
ivelten, line 294: I don't like this folding design, but it works as expected for now. I think I will try to make a better design in future.

ivelten:
Moved to #203.

PR #202, opened 2018-12-13
Stream batching

ivelten:
With this pull request, I am implementing a way to produce buffers in streamed queries, so the user can batch results by a window of time.


ivelten:
Moved to #203.

PR #212, opened 2019-02-01
Adjusting tests for machines with less processor cores

ivelten:
This PR is meant to solve the inconsistency in test runs made against machines with less cores - specially CI ones (which normally run with 2 cores).


johnberzy-bazinga:
LGTM 👍

PR #225, opened 2019-05-17
Fix bug where Resolvers were not being run in parallel, and refactor Execution.fs

TOTBWF:
This PR fixes a few major bugs:
1) Resolvers are not being run in parallel. This can cause 10x slowdowns for even
very simple workloads.
2) Each deferred/streamed field would cause the entire query to be re-run.

This PR also refactors `Execution.fs`, and cleans up a lot of the ad-hoc error handling/defer stuff.


TOTBWF:
There are still a few bugs to be worked out around live execution. As it stands, the API seems a bit clunky. For example, this is how the API looks now:
```
let sub =
    { FieldName = "live"
      TypeName = "Data"
      Identity = fun (x : TestSubject) -> x.id }

schemaConfig.LiveFieldSubscriptionProvider.Register sub

let updateLiveData value=
    data.live <- value
    schemaConfig.LiveFieldSubscriptionProvider.Publish "Data" "live" data
```

Instead, why don't we do something similar to `SubscriptionFields`?
```
type Foo = { Id : string, mutable live : int }
let LiveField : LiveFieldDef<Foo> = Define.LiveField(
    "live",
    String,
    fun ctx oldValue newValue -> if oldValue.Id = newValue.Id then Some (newValue.int) else None)
```

TOTBWF:
Furthermore, we should probably rethink how to handle buffered streams.
The result format we use now is
```json
{
    "data" : [
        { "value" : 1 }
        { "value" : 2 }
    ],
    "path" : [ "foo", "bar", [1, 2] ]
}
```

While this does de-duplicate the path, it also means that we lose a consistent way of handling buffered and non-buffered streams.

TOTBWF:
There are also some questions that need to be answered when dealing with subscriptions and live/stream/defer. For example, consider the query
```graphql
subscription StoryLikeSubscription($input: StoryLikeSubscribeInput) {
  storyLikeSubscribe(input: $input) {
    story {
      likers @stream { count }
      likeSentence { text @live }
    }
  }
}
```
Should these get streamed in? Furthermore, how do we know which subscription result to patch the deferred bits into?

PR #231, opened 2019-06-12
Fixing heading anchor and small typo

miighty-maus:


johnberzy-bazinga:
@miighty-maus Thanks!

PR #233, opened 2019-06-28
Parser updates

ivelten:
With this PR, I am proposing updates to the parser. It needs support for two features of GraphQL:

- [x] Correctly parsing `null` keyword into a [Null Value](https://graphql.github.io/graphql-spec/June2018/#sec-Null-Value). It actually parses the keyword into an Enum Value.
- [x] Correctly parsing an unnamed operation. The parser does recognize short-hand queries, like `{ field }`, but does not recognize unnamed queries with the operation type keyword, like `query { field }`.


johnberzy-bazinga:
LGTM 👍

PR #234, opened 2019-07-02
Client provider subscription support

ivelten:
With this PR, I am proposing a subscription interface for the client provider.

The proposal specifies that we should try to be most agnostic possible from subscription protocols, in case the client code needs to connect to a custom implementation. We offer [this spec](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) as the default one, in case no one is provided, though.

- [x] Create an interface to represent a subscription handler, with all sub-interfaces needed
- [x] Provide a default implementation based on the [GraphQL Over Web Socket specification](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)
- [ ] Integrate the interface and the default implementation on the provider
- [ ] Integration tests
- [ ] Code review and refactoring


ivelten:
Closing this as it will be opened in another PR.

PR #235, opened 2019-07-02
Client provider subscription support

ivelten:
With this PR, I am proposing a subscription interface for the client provider.

The proposal specifies that we should try to be most agnostic possible from subscription protocols, in case the client code needs to connect to a custom implementation. We offer [this spec](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) as the default one, in case no one is provided, though.

- [x] Create an interface to represent a subscription handler, with all sub-interfaces needed
- [x] Provide a default implementation based on the [GraphQL Over Web Socket specification](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)
- [ ] Integrate the interface and the default implementation on the provider
- [ ] Integration tests
- [ ] Code review and refactoring


src/FSharp.Data.GraphQL.Client/Subscriptions.fs
```diff
@@ -0,0 +1,247 @@
...
+        member __.Data = data
+        member __.Deferred = deferred
+        member __.Unsubscribe() = unsubscriber () |> Async.RunSynchronously
+        member __.AsyncUnsubscribe() = unsubscriber ()
+
+    interface IDisposable with
+        member __.Dispose() = unsubscriber () |> Async.RunSynchronously
+
+/// An interface for implementing subscription handlers for GraphQLProvider.
+type IGraphQLSubscriptionHandler =
```
ivelten, line 144: This is a standard interface for implementing subscription handlers, in case the client code does not use the GraphQL Over Web Socket spec that we default to.

src/FSharp.Data.GraphQL.Client/Subscriptions.fs
```diff
@@ -0,0 +1,247 @@
...
+    /// When implemented, start a subscription to the GraphQL using the implementation protocol.
+    abstract member Subscribe : GraphQLSubscriptionRequest -> IGraphQLSubscriptionResponse
+    /// When implemented, closes the subscription connection to the GraphQL server using the implementation protocol.
+    abstract member Close : unit -> unit
+    /// When implemented, asynchronously closes the subscription connection to the GraphQL server using the implementation protocol.
+    abstract member AsyncClose : unit -> Async<unit>
+
+/// A GraphQL subscription handler which uses GraphQL Over Web Socket protocol.
+/// See https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md for information.
+type GraphQLOverWebSocketSubscriptionHandler() =
```
ivelten, line 161: This is the default implementation of the handler, which uses the spec we default to. The idea is to share it with each subscription call of the provider because this connection can be reused like the classical HTTP connection.

src/FSharp.Data.GraphQL.Client/Subscriptions.fs
```diff
@@ -0,0 +1,247 @@
...
+        |> tryPickFirstAndKeepRemaining
+        |> Option.map (fun (first, remaining) ->
+            let deferred = {
+                new IObservable<JsonValue> with
+                    member __.Subscribe(observer) =
+                        remaining
+                        |> Seq.iter (function
+                            | Incoming payload -> observer.OnNext(payload)
+                            | Finished -> observer.OnCompleted())
+                        { new IDisposable with member __.Dispose() = () } }
```
ivelten, line 128: We pick the first item from the response and use it as the direct result.
The remaining items will be processed into the deferred result as an `IObservable`.
Every Observable returns `JsonValue` here. They will be mapped into provided types in the `ProvidedTypesHelper`.

src/FSharp.Data.GraphQL.Client/Subscriptions.fs
```diff
@@ -0,0 +1,247 @@
...
+        |> Option.defaultValue (None, Observable.empty)
+
+    interface IGraphQLSubscriptionResponse with
+        member __.Data = data
+        member __.Deferred = deferred
+        member __.Unsubscribe() = unsubscriber () |> Async.RunSynchronously
+        member __.AsyncUnsubscribe() = unsubscriber ()
+
+    interface IDisposable with
+        member __.Dispose() = unsubscriber () |> Async.RunSynchronously
```
ivelten, line 141: Unsubscriber, in this case, will send a message to the server to end the connection (`Stop <id>`). When the server responds with `Complete <id>`, it will be processed in the observable above and call `OnComplete` to properly end the observer subscription.

xperiandri:
Is it already implemented in the client?

xperiandri:
@valbers I think we need to extract your server-side models, JSON serializer and common logic (if applicable) to a separate project `FSharp.Data.GraphQL.WebSockets` so that it is used in `FSharp.Data.GraphQL.Server.AspNetCore`, `FSharp.Data.GraphQL.Client` and `Snowflaqe`

valbers:
Just for the record: the changes in this PR are based on the superseded websockets subprotocol `subscriptions-transport-ws`. The current version of the protocol is `graphql-ws` (f.k.a. `graphql-transport-ws`). Our server-side subscriptions implementation over websockets already implements `graphql-ws`.

PR #239, opened 2019-07-19
Orleans Query Tree Middleware proposal

ivelten:
With this pull request, I am proposing a middleware for serialization of a query tree. The initial idea for it is to send this query tree with middleware to some destination, like Microsoft Orleans.


johnberzy-bazinga:
we're going to implement a different solution to serializing the AST that isn't tied to Orleans.

PR #259, opened 2020-01-28
Traverse list values in argument lists

hth313:
Validating usage of variables did not traverse list values in
arguments, resulting in incorrect validation errors.

Also add a test case for it.


hth313:
This fixes https://github.com/fsprojects/FSharp.Data.GraphQL/issues/253

PR #264, opened 2020-03-09
Remove GET request

Birnsen:
With some GraphQL Servers it is possible that
the GET request does not fail and a proper
message is returned, although this message is
an error message. For this case the typeprovider
fails to get the introspection information.

This commit removes the GET Request altogether
and uses just the POST Request method, which is
the common standard I suppose.

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/Ast.fs
+++ b/src/FSharp.Data.GraphQL.Shared/Ast.fs
@@ -26,13 +26,16 @@ and Definition =
         | FragmentDefinition frag -> frag.SelectionSet
 
 /// 2.2.1 Operations
-and OperationDefinition = {
-    OperationType: OperationType
-    Name: string option
-    VariableDefinitions: VariableDefinition list 
-    Directives: Directive list
-    SelectionSet: Selection list
-}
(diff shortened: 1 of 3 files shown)


Birnsen:
This is my first Pull Request, so if I have done anything wrong please advice me to correct any issues :).

In case this is not wanted in that way I have another solution in fixing this issue, which is a little bit more involved.

Birnsen:
Do I have to rework this one? The failing CI seems to be because of missing FSharp.Core. At least I think that this is the problem!?

johnberzy-bazinga:
> This commit removes the GET Request alltogether
and uses just the POST Request method, which is
the common standard.

Can you provide some references to tools that use a get for the introspection schema? I think most tooling actually submits a query, while our client expects a static json response. 

Since this could potentially break some existing clients a safer approach would be to validate the response is actually a valid schema and only falling back to a post if necessary. 

Thanks,
John

Birnsen:
Yeah I think your´re right if it breaks other clients the proper approach is to validate the response and fall back to Post if necessary. I will alter this PR in the next couple days.

PR #265, opened 2020-03-09
Enable error messages without a path

Birnsen:
Not all error messages sent back from a
GraphQL Server contain a path.

This commit adds an additional case for error
messages without a path. The path is set to
an empty array for the returned type.


johnberzy-bazinga:
Thanks for the contribution!

PR #273, opened 2020-03-28
Fix introspection for Servers expecting Post

Birnsen:
It is possible that a Get Request is successful
but does not contain a proper schema.

With this commit if a Get Requests succeds it
will be checked if a schema can be deserialized.
If not a Post request is taking place and also
checked for a proper schema.


src/FSharp.Data.GraphQL.Client/GraphQLClient.fs
```diff
@@ -88,7 +88,7 @@ module GraphQLClient =
         |> Async.RunSynchronously
 
     /// Executes an introspection schema request to a GraphQL server asynchronously.
-    let sendIntrospectionRequestAsync (connection : GraphQLClientConnection) (serverUrl : string) httpHeaders =
+    let getIntrospectionAsync (connection : GraphQLClientConnection) (serverUrl : string) httpHeaders =
         let sendGet() = async { return! getAsync connection.Client serverUrl }
```
xperiandri: Why do we need to wrap that?
```suggestion
        let sendGetAsync() = getAsync connection.Client serverUrl
```

xperiandri (changes requested):

xperiandri (approved):

xperiandri:
Could you rebase and solve conflicts, please?

Birnsen:
I currently have no access to a Dev-Environmnt and the PR is rather old, so I do not know if the Rebase worked correctly.

xperiandri:
This is handled in the latest versions. Thanks for contribution!

PR #278, opened 2020-04-03
Fixed issue with HttpClient creation for each request

xperiandri:
Also updated ClientProvider to use HttpMessageInvoker (HttpMessageHandler) instead of server URL and HTTP headers separately.
And switched to use `Stream` from `HttpResponse` instead of reading to a `string` immediately.

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -7,7 +7,7 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
 
     <DetectedMSBuildVersion>$(MSBuildVersion)</DetectedMSBuildVersion>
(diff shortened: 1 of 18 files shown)


xperiandri:
@johnberzy-bazinga, could you have a look and let me know if it is possible to overcome `error FS3022: The type provider 'FSharp.Data.GraphQL.Client.GraphQLTypeProvider' reported an error in the context of provided type 'FSharp.Data.GraphQL.GraphQLProvider,introspection="http://localhost:8085",uploadInputTypeName="Upload"+Operations+Q', member 'Run'. The error: Unsupported constant type 'FSharp.Data.GraphQL.GraphQLRuntimeContextInfo'. Quotations provided by type providers can only contain simple constants. The implementation of the type provider may need to be adjusted by moving a value declared outside a provided quotation literal to be a 'let' binding inside the quotation literal.` limitation?

PR #279, opened 2020-04-04
Fixed issue with HttpClient creation for each request

xperiandri:
Also
- updated ClientProvider to use HttpMessageInvoker (HttpMessageHandler).
- switched to use Stream from HttpResponse instead of reading to a string immediately.
- added `GetContext` overload that takes `HttpClient` factory


xperiandri:
@johnberzy-bazinga, client provider builds, but for operations I get `error FS3025: Unsupported expression 'FieldGet (None, Empty)' from type provider. If you are the author of this type provider, consider adjusting it to provide a different provided expression.`
What can I broke?

PR #284, opened 2020-04-18
Added .editorconfig

xperiandri:


xperiandri:
@ivelten, @johnberzy-bazinga do you agree with these changes?

ivelten:
@xperiandri I like to use this exact style in my .editorconfig files for .NET solutions. I think we could have this file to help maintain code formatting standards.

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+insert_final_newline     = true
+trim_trailing_whitespace = true
+# Project files
+[*.{csproj,fsproj}]
+indent_size = 2
+charset     = utf-8-bom
+# Code files
+[*.{cs,csx,fs,fsx,vb,vbx,xml}]
+indent_size = 4
+charset     = utf-8-bom
```
aggieben: Why BOM instead of just utf-8?

xperiandri: Usually, I've seen this option set to BOM

xperiandri:
Close in favor of #323

xperiandri:
@ivelten, @mickhansen, @nikhedonia

1. What do you think about the proposed settings?
2. Do you prefer the first or the second commit settings?
3. Would you like anything to add? Maybe about JavaScript

xperiandri:
Now the right commits are in the right place

aggieben:
This seems fine to me 👍 (with the caveat that I can't fully visualize what everything would look like just from the settings).

xperiandri:
Maybe you have your favourite preset 🙂

xperiandri:
To be honest I have not checked for recent new settings in Fantomas. So maybe I miss some of them

mickhansen:
@xperiandri I have no strong feelings about this :)

njlr:
Perhaps we should copy FSharp itself? https://github.com/dotnet/fsharp/blob/main/.editorconfig

xperiandri:
@njlr good idea. I had a look into into it and borrowed 1 settings

xperiandri:
So the latest commit has settings aligned with the latest Fantomas config ([see](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#configuration))

xperiandri:
If no objections are provided today tomorrow I will update all the files according to config

.editorconfig
```diff
@@ -0,0 +1,57 @@
...
+# Project files
+[*.{csproj,fsproj,vbproj,shproj,projitems}]
+indent_size          = 2
+charset              = utf-8-bom
+
+###############################
+# F# Coding Conventions       #
+###############################
+[*.{fs,fsx}]
+max_line_length = 120
```
nojaf: Please don't add all the possible Fantomas settings.
This is not recommended, https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#configuration

> Please note that you should only add settings to the .editorconfig file when you want to deviate from the default settings. Copying the entire list above is unnecessary.

xperiandri: Thank you for the suggestion!

PR #285, opened 2020-04-21
Allowing GraphQLClientConnection to be reusable between requests

ivelten:
With this Pull Request, I am proposing a simple working solution for the anti-pattern of creating a new `HttpClient` instance for each `GraphQLClientConnection` usage by the Provider.

First, `GetContext` method of the provider now allows the user to pass an instance of `unit -> GraphQLClientConnection` factory, so the user can control its connection lifecycle the way he wants to.

I think the best approach would be remove the `GraphQLClientConnection` wrapper around the `HttpClient`, but the `HttpClient` has some serious reference issues with Mono, and just trying to make it exposed in design time gives a lot of errors that I could not solve, so I kept it wrapped inside the client connection, using it only internally.

I was able to make a new overload of the `GraphQLClientConnection` constructor so it can receive a new instance of the `HttpClient`, and also allowing the user to decide if the wrapper should owns the connection (disposing it when it is disposed) or not.

I avoided creating a Singleton to cache the `HttpClient` because of [potential issues with DNS changes](https://stackoverflow.com/a/48778707). I think it is best for the client user himself to manage these situations.

UPDATE: the `GraphQLClientConnection` now uses `HttpMessageInvoker` instead of `HttpClient`.

### DIFF
--- a/src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
+++ b/src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
@@ -9,15 +9,22 @@ open System.Net.Http
 /// The connection component for GraphQL client operations.
 [<AllowNullLiteral>]
 type GraphQLClientConnection(invoker : HttpMessageInvoker, ownsInvoker : bool) =
+    do if isNull invoker then raise <| ArgumentNullException("invoker")
+
     /// Creates a new instance of GraphQLClientConnection, using a provided System.Net.Http.HttpMessageInvoker to be wrapped inside.
     /// The provided System.Net.Http.HttpMessageInvoker is not owned by the connection, so it must be disposed when it is not necessary anymore.
     new(invoker) = new GraphQLClientConnection(invoker, false)
+
+    /// Creates a new instance of GraphQLClientConnection, using a provided System.Net.Http.HttpClient as a System.Net.Http.HttpMessageInvoker to be wrapped inside.
+    /// The provided System.Net.Http.HttpMessageInvoker is not owned by the connection, so it must be disposed when it is not necessary anymore.
+    new(invoker : HttpClient) = new GraphQLClientConnection(invoker, false)
 
     /// Creates a new instance of GraphQLClientConnection, using a fresh new instance of System.Net.Http.HttpClient as an System.Net.Http.HttpMessageInvoker.
     /// This instance is owned by the GraphQLClientConnection, being disposed when the connection itself is disposed.
     new() = new GraphQLClientConnection(new HttpClient(), true)
 
-    member internal __.Invoker = invoker
+    /// The System.Net.Http.HttpMessageInvoker used by this connection.
+    member __.Invoker = invoker
     
     interface IDisposable with
         member __.Dispose() = if ownsInvoker then invoker.Dispose()
--- a/src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
+++ b/src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
@@ -9,8 +9,12 @@ open System.Net.Http
 /// The connection component for GraphQL client operations.
 [<AllowNullLiteral>]
 type GraphQLClientConnection(client : HttpClient, ownsClient : bool) =
+    /// Creates a new instance of GraphQLClientConnection, using a provided HttpClient to be wrapped inside.
+    /// The provided HttpClient is not owned by the connection, so it must be disposed when it is not necessary anymore.
     new(client) = new GraphQLClientConnection(client, false)
 
+    /// Creates a new instance of GraphQLClientConnection, using a fresh new instance of HttpClient internally.
+    /// This instance is owned by the GraphQLClientConnection, being disposed when the connection itself is disposed.
     new() = new GraphQLClientConnection(new HttpClient(), true)
 
     member internal __.Client = client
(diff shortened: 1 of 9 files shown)


src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
```diff
@@ -0,0 +1,19 @@
...
+/// Copyright (c) 2016 Bazinga Technologies Inc
+
+namespace FSharp.Data.GraphQL
+
+open System
+open System.Net.Http
+
+/// The connection component for GraphQL client operations.
+[<AllowNullLiteral>]
+type GraphQLClientConnection(client : HttpClient, ownsClient : bool) =
```
xperiandri: Will it work making it `struct`?

xperiandri:
Looks very good. I'm satisfied with this solution. Thank you

xperiandri:
As a separate PR I would also move to from using `HttpClient` to using its parent `HttpMessageHandler`

src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
ivelten: Well, we can. Will it be better as a `struct`?

ivelten:
@xperiandri I think by parent you mean `HttpMessageInvoker`, right? I tried using it, but `HttpMessageInvoker` does not have an implementation of `GetAsync`, which is used to introspect the server schema. So I used `HttpClient` instead.

ivelten:
Actually, never mind. I totally forgot that I can make a GET request using `SendAsync` method. Will update the PR soon.

src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
xperiandri: Yes, it is a single field. For performance, a struct is better

src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
```diff
@@ -8,16 +8,16 @@ open System.Net.Http
...
 /// The connection component for GraphQL client operations.
 [<AllowNullLiteral>]
-type GraphQLClientConnection(client : HttpClient, ownsClient : bool) =
-    /// Creates a new instance of GraphQLClientConnection, using a provided HttpClient to be wrapped inside.
-    /// The provided HttpClient is not owned by the connection, so it must be disposed when it is not necessary anymore.
-    new(client) = new GraphQLClientConnection(client, false)
+type GraphQLClientConnection(invoker : HttpMessageInvoker, ownsInvoker : bool) =
+    /// Creates a new instance of GraphQLClientConnection, using a provided System.Net.Http.HttpMessageInvoker to be wrapped inside.
+    /// The provided System.Net.Http.HttpMessageInvoker is not owned by the connection, so it must be disposed when it is not necessary anymore.
+    new(invoker) = new GraphQLClientConnection(invoker, false)
```
xperiandri, line 16: Could you also add a constructor that accept `HttpClient` so that library user does not need to upcast

src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
ivelten: The problem with making it a `struct` is that it does have a nullable value inside, and with a forced default constructor that does not initialize it, we can have a `GraphQLClientConnection` that actually holds no `HttpMessageInvoker`. I don't feel secure keeping a reference type as a value inside a structure. Do you think we should do it anyway? Any suggestion?

src/FSharp.Data.GraphQL.Client/GraphQLClientConnection.fs
ivelten: Sure.

xperiandri:
@ivelten, @johnberzy-bazinga is it ready to merge?
From the point of my request, it looks ready.

ivelten:
I guess I have no more points for now. Looks good to me too.

xperiandri:
@johnberzy-bazinga, let's merge?

aggieben:
Just as a point of fact for the record for later, .NET Core 2.1 introduced the [`IHttpClientFactory`](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests) to deal with this issue.  In a future where we aren't multitargeting we might want to consider making use of it.

johnberzy-bazinga:
Thanks @ivelten and @xperiandri

xperiandri:
@aggiebe, how would you use `IHttpClientFactory` for GraphQL client?
As I see it is the responsibility of a consumer to pass the right HTTP client. GraphQL client has no background to choose a particular client and the use of it is primarily limited to `Microsoft.Extensions.DependencyInjection`

aggieben:
@xperiandri I would imagine you'd have the user pass in an `IHttpClientFactory`, rather than an `HttpMessageInvoker`.

xperiandri:
@aggieben / @johnberzy-bazinga could you issue a new version of NuGet with these changes?

xperiandri:
@jberzy, @ivelten, please, issue a new client NuGet package
I'm going to present the library at the local conference. I need this functionality

ivelten:
I could help making a new Pull Request with those changes, but I can only work on it next weekend. If that's okay, we could talk about this.

xperiandri:
Yes, my conference talk is at the beginning of December

ivelten:
@xperiandri I could not focus on this weekend, I will be pushing a PR sometime during this week.

ivelten:
Sorry for the long delay, @xperiandri. Here is the [PR](https://github.com/fsprojects/FSharp.Data.GraphQL/pull/309).

Issue #290, opened 2020-06-20
Server architecture for large schemas. Type extensions?


agu-z:
My team is migrating an existing GraphQL API from JS to F#. Our API has a lot of types, so we decided to create a file per type (Define.Object). 

The file structure looks like this:
- Schema
  - User.fs
  - Organization.fs
  - ...
  - Query.fs

If `Organization` has a `users` field that returns `ListOf User`, you can simply compile `User.fs` before `Organization.fs`. 
However, this doesn't work if `User` also needs to have an `organizations` field, because F# doesn't allow mutual recursion across files.

What is the recommended way to architecture a big server like this? The samples use a single schema file, but unfortunately, that's just not feasible for our use case. If we did this, we would eventually have a file with 20K+ lines of code (judging by the size of our JS API).

A solution might be supporting Type Extensions. Especially, Object Extensions: http://spec.graphql.org/June2018/#sec-Object-Extensions

With Object Extensions, we would be able to define these relationship fields at the return type module instead. In my example, `Organization.fs` would add the `organizations` field to the `User` type, and `User.fs` would add the `users` field to the `Organization` type.

This architecture would work out pretty nicely for us because the querying and authorization logic needed to return records of a type would remain in its module as opposed to being spread across the codebase. 

Supporting GraphQL extensions would also allow users of this library to build GraphQL microservices, that could be composed using tools like [Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/).

If you think that's a good idea, I'd be more than happy to contribute. Thanks!

johnberzy-bazinga:
@agu-z  My apologies for the late reply. As you alluded to we don't have a good solution for large schemas at present. We would be more than happy to take a contribution. Do you have a bit of time to sketch out what the API would look like for type extensions? 

Presumably, The User and Organization records would be defined at top level e.g. `Domain.fs` and only the GraphQL Schema related code resides in `Organization.fs` and `User.fs` (in order to make the extensions typed).  Am I getting that right? 

Thanks,
John

agu-z:
Hey @johnberzy-bazinga, don't apologize! Thank you for getting back to me :)

[...]

PR #294, opened 2020-07-19
ObjectRef implementation

agu-z:
This my attempt at implementing an `ObjectRef` function in an effort to alleviate some of the pain-points of defining large schemas using this library. This approach was proposed by @johnberzy-bazinga on #290. 

This a work-in-progress implementation and it will currently fail with nested references. I'm leaving some questions on how to proceed.

Thanks,
Agus

### DIFF
--- a/samples/star-wars-api/Schema.fs
+++ b/samples/star-wars-api/Schema.fs
@@ -191,7 +191,8 @@ module Schema =
             fields = [
                 Define.Field("hero", Nullable HumanType, "Gets human hero", [ Define.Input("id", String) ], fun ctx _ -> getHuman (ctx.Arg("id")))
                 Define.Field("droid", Nullable DroidType, "Gets droid", [ Define.Input("id", String) ], fun ctx _ -> getDroid (ctx.Arg("id")))
-                Define.Field("planet", Nullable PlanetType, "Gets planet", [ Define.Input("id", String) ], fun ctx _ -> getPlanet (ctx.Arg("id")))
+                // TODO: Create another sample for ObjectRef
+                Define.Field("planet", Nullable (Define.ObjectRef<Planet>("Planet")), "Gets planet", [ Define.Input("id", String) ], fun ctx _ -> getPlanet (ctx.Arg("id")))
                 Define.Field("characters", ListOf CharacterType, "Gets characters", fun _ _ -> characters) ])
 
     let Subscription =
@@ -206,7 +207,7 @@ module Schema =
                     [ Define.Input("id", String) ],
                     (fun ctx _ p -> if ctx.Arg("id") = p.Id then Some p else None)) ])
 
-    let schemaConfig = SchemaConfig.Default
+    let schemaConfig = { SchemaConfig.Default with Types = [PlanetType]}
 
     let Mutation =
         Define.Object<Root>(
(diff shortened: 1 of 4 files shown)


samples/star-wars-api/Schema.fs
```diff
@@ -191,7 +191,8 @@ module Schema =
             fields = [
                 Define.Field("hero", Nullable HumanType, "Gets human hero", [ Define.Input("id", String) ], fun ctx _ -> getHuman (ctx.Arg("id")))
                 Define.Field("droid", Nullable DroidType, "Gets droid", [ Define.Input("id", String) ], fun ctx _ -> getDroid (ctx.Arg("id")))
-                Define.Field("planet", Nullable PlanetType, "Gets planet", [ Define.Input("id", String) ], fun ctx _ -> getPlanet (ctx.Arg("id")))
+                // TODO: Create another sample for ObjectRef
```
agu-z, line 194: I'll provide better documentation and tests once we settle on the approach.

src/FSharp.Data.GraphQL.Server/Schema.fs
```diff
@@ -177,6 +177,41 @@ type Schema<'Root> (query: ObjectDef<'Root>, ?mutation: ObjectDef<'Root>, ?subsc
...
+    and resolveObjectReferences (objDef: ObjectDef<'a>): ObjectDef<'a> =
+        objDef.WithFields <|
+            Seq.map (fun (KeyValue (_, field)) ->
+                upcast CustomOutputFieldDefinition(field, downcast resolveTypeReference field.TypeDef)
+            ) objDef.Fields
+
+        
+        
+    let query = resolveObjectReferences query
+    let mutation = Option.map resolveObjectReferences mutation
```
agu-z, line 214: @johnberzy-bazinga As you can see, the way I went about this is by creating a new `ObjectRef` type that's resolved to real Object Definition at Schema instantiation. I also tinkered with the idea of resolving these references in the Execution Phase, but I thought that could be more error-prone. Let me know if the approach I've taken makes sense to you.

agu-z: Also, this only works when using references at the Root level because it doesn't go into nested Objects. 
Check out line 194. Downcasting `objDef` fails, saying it cannot cast `ObjectDefinition<Planet>` to `ObjectDef<obj>`. 

Any ideas on how to deal with this case? 

I think I'm using the wrong types in one of these functions, but I wanted to ask you before attempting something too hacky.

jberzy: I think it makes sense before the execution step, but a couple of points:

1) The TypeMap.ofSeq call above traverses the tree and I think it will fail unless we [treat the ObjectRef case](https://github.com/fsprojects/FSharp.Data.GraphQL/blob/dev/src/FSharp.Data.GraphQL.Shared/TypeSystem.fs#L1881).
2) I believe We'll need to readd the ObjectDefs to the TypeMap after (or maybe during?) the `resolveObjectReferences` step since we're modifying the fields.

jberzy: @agu-z  Regarding the exception for casting, I think there is no way around that without using reflection. We need to call `resolveObjectReferences<'a>` using reflection where `'a` is a `Planet` in the specific case.

Alternatively, we can look to add the untyped equivalent of `ObjectDef<'Val>.WithFields` and work with that instead in `resolveObjectReferences`.

jberzy: I think the general approach is right here. If we do it early on in the Schema initialization, we don't need to worry about dealing with ObjectRefs in later phases.

agu-z: Makes sense to me. I apologize for the radio silence. A lot has changed at work and I have been terribly busy. I'll try to take some time next week to finish up this PR.

xperiandri:
@agu-z any intent to complete this PR?

PR #312, opened 2020-12-18
Reduce constructors to one: Use optional arguments (wrapped in Option).

### LINKED ISSUE #250: Question: Why overloads instead of optional parameters?
Edit:

I'm giving the FSharp.Data.GraphQL client provider another try, since it's been a few months.  Looks like I'm not running into the same errors I got last time.

One thing I notice is that when I create an operation using a query like so:

```fsharp
let op = B.Operation<"""            query ($uId:Int, $rId:Int, $sinceTimestamp:String, $first:Int, $offset:Int) {
              answer(id: $rId) {
                agreed_users (sinceTimestamp:$sinceTimestamp, first:$first, offset:$offset) {
                    id
                    is_verified
                    username
                    avatar
                    follows_status(uId: $uId)
                }
              }
            }""">()
```

`op.Run` creates an overload for every permutation of parameters, plus a runtime context.

Is there any reason [optional parameters](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/parameters-and-arguments#optional-parameters) were not used instead?  Seems like it would be much easier to read and use, and probably more performant too.  If there's no particular reason, is that something you would consider?

bddckr:
🚨 This PR comes with a breaking change! 🚨

This PR changes the constructor count for record types generated by the type provider to one. This single constructor honors the GraphQL schema by requiring the required parameters and uses optional parameters to set the rest.

As outlined in both the source code before this change, but also in #250, this change requires something one could classify as a workaround of a shortcoming of the F# type provider SDK. This PR resolves the issue in a way that the single constructor can be used in the following manner:
1. Not providing a default parameter results in that parameter _not_ getting included in the request.
2. If `None` or `?parameterName = Some None` is used it has the same effect as bullet point one.
3. To specify an actual value to be sent with the request one has to provide `Some value` (or, if named arguments are required/preferred: `parameterName = Some value`).

Additionally there's something to be aware of:
At least Visual Studio 2019 (with the .NET 5 SDK used) is showing the optional parameters as `[?parameterName TypeName]` in IntelliSense. This is misleading, as shown when hovering over the named argument: The parameter is actually of type `TypeName option` and since it's optional that leads to `(TypeName option) option`. The compiler does not have this issue and adds the necessary diagnostics in case the passed value is of the wrong type.
This explains the above required usage notes.

The _required_ parameters behave as expected: There's no IDE issue, nor any specific usage notes to be aware of.

Resolves #250.

---

I'm opening this PR as a draft for discussion and am asking for feedback. I'm happy to discuss anything raised! Here's what I believe needs to be discussed and ultimately decided on:
1. Is this breaking change of removing all constructor overloads and only keeping this single constructor OK?
	I'm aware that there's a suggestion in #250 to make this constructor optional, by keeping the existing ones (and/or making this behavior configurable for each type provider instance). However, I believe that this solution is the "sweet spot":
	Users don't have to search through all permutations of possible constructor parameter combinations (>500 in my case) in their IDE's IntelliSense/auto-completion, they immediately are greeted by the single contract that is defined by the GraphQL schema.
2. The "usage notes" above seem complex, but I believe it's perfectly fine as is. Users are just able to be explicit about "include this value" and "don't send this value at all". Do we agree? 😄
3. Tests... I'll defer updating the existing ones until we decide on whether to replace all constructors with this single one or add it etc.

### DIFF
--- a/src/FSharp.Data.GraphQL.Client.DesignTime/DesignTimeCache.fs
+++ b/src/FSharp.Data.GraphQL.Client.DesignTime/DesignTimeCache.fs
@@ -15,7 +15,8 @@ type internal ProviderKey =
       CustomHttpHeadersLocation : StringLocation
       UploadInputTypeName : string option
       ResolutionFolder : string
-      ClientQueryValidation: bool }
+      ClientQueryValidation: bool
+      ExplicitOptionalParameters: bool }
 
 module internal ProviderDesignTimeCache =
     let private expiration = CacheExpirationPolicy.SlidingExpiration(TimeSpan.FromSeconds 30.0)
--- a/src/FSharp.Data.GraphQL.Client.DesignTime/ProvidedTypesHelper.fs
+++ b/src/FSharp.Data.GraphQL.Client.DesignTime/ProvidedTypesHelper.fs
@@ -138,7 +138,7 @@ module internal Failures =
 module internal ProvidedRecord =
     let ctor = typeof<RecordBase>.GetConstructors().[0]
 
-    let makeProvidedType(tdef : ProvidedRecordTypeDefinition, properties : RecordPropertyMetadata list) =
+    let makeProvidedType(tdef : ProvidedRecordTypeDefinition, properties : RecordPropertyMetadata list, explicitOptionalParameters: bool) =
         let name = tdef.Name
         tdef.AddMembersDelayed(fun _ -> 
             properties |> List.map (fun metadata ->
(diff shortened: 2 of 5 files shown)


jberzy:
@bddckr Thanks for the PR! I agree that this is probably the "sweet spot", but I'm just concerned about a breaking change in a minor release. What do you think about parameterizing the behaviour so we don't have to worry about breaking peoples apps? We can make this the default behaviour in the next major release. 

Also, I've seen you applied this change only to InputObject constructors. What do you think about also applying this to method calls with optional arguments?

Thanks,
John

bddckr:
> What do you think about parameterizing the behaviour so we don't have to worry about breaking peoples apps? We can make this the default behaviour in the next major release.

That makes sense. I'll have a look at adding an optional parameter to the type provider then. I'll use that suggested name from the issue (`explicitOptionalParameters`). In the next major version we can then change the default argument's value to be the other way round. 👍

> Also, I've seen you applied this change only to InputObject constructors. What do you think about also applying this to method calls with optional parameters?

Ah I knew I forgot something! 😅 I'll definitely add that, too. This will also be behind the above new boolean flag to toggle this backwards-compatibility breaking behavior then. Oh and I'll follow what I did here: _Replace_ all existing overload permutations with a _single_ one with optional parameters (if the flag is turned on).


Does that all sound like it should be OK?

bddckr:
Done both of those suggestions now 👍 I think this is ready for review now.

I don't see any tests for the type provider, but let me know if there are any to which I can add some more for this PR!

jberzy:
Hi @bddckr thanks for the updates to this PR. Looks good! In terms of tests, do you mind adding a couple of tests to cover both input object and run methods with optional parameters. You can find the tests under https://github.com/fsprojects/FSharp.Data.GraphQL/tree/dev/tests/FSharp.Data.GraphQL.IntegrationTests

bddckr:
Oh I have totally missed the integration tests because I forgot that there's a second sln. 😅 Thanks for the help!

I've duplicated the existing local provider tests so we have the whole batch running against a provider with the optional parameters turned on. I tried to have less tests, but the more I wrote the more obvious it became that it's probably a better idea to test all scenarios: All of them are different due to the constructor/method overloads being down to one (or two if you count the ones that allow passing in a context) now.
I'm afraid I'll need some hints as to what to test specifically otherwise. I know it's probably not in the interest to repeat all local provider tests for this change, but then again we said we want to make this the default setting down the road anyway...

jberzy:
@bddckr LGTM! Thanks!

bddckr:
@jberzy Thanks for the help and merging it! Will this be available on NuGet soon?

I noticed the release notes document is missing the last release. It should probably also be updated to state this change once that rolls out. 😄

jberzy:
I've published a new nuget (1.0.7) and update the release notes. Thanks for the contribution!

PR #329, opened 2021-04-06
bugfix/parser-escaped-strings

njlr:
Small change, just needed to flip around `normalCharacter <|> escapedCharacter` to `escapedCharacter <|> normalCharacter`. I also added a unit test.


jberzy:
Thanks for the PR

PR #339, opened 2021-06-08
add Path to ResolveFieldContext

mickhansen:
This PR adds `Path` to `ResolveFieldContext` to make it easier to make pretty errors when adding custom error handling (with custom exceptions) via middleware and `ctx.AddError`

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -224,7 +224,7 @@ let private resolveUnionType possibleTypesFn (uniondef: UnionDef) =
     | Some resolveType -> resolveType
     | None -> defaultResolveType possibleTypesFn uniondef
 
-let private createFieldContext objdef argDefs ctx (info: ExecutionInfo) =
+let private createFieldContext objdef argDefs ctx (info: ExecutionInfo) (path : obj list) =
     let fdef = info.Definition
     let args = getArgumentValues argDefs info.Ast.Arguments ctx.Variables
     { ExecutionInfo = info
@@ -233,7 +233,8 @@ let private createFieldContext objdef argDefs ctx (info: ExecutionInfo) =
       ParentType = objdef
       Schema = ctx.Schema
       Args = args
-      Variables = ctx.Variables }
+      Variables = ctx.Variables
+      Path = path }
 
 let private resolveField (execute: ExecuteField) (ctx: ResolveFieldContext) (parentValue: obj) =
     if ctx.ExecutionInfo.IsNullable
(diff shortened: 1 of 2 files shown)


src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -878,7 +878,9 @@ and ResolveFieldContext =
       /// parametrized inputs.
       Args : Map<string, obj>
       /// Variables provided by the operation caller.
-      Variables : Map<string, obj> }
+      Variables : Map<string, obj>
+      /// Field path
+      Path : obj list }
```
jberzy, line 883: Hi @mickhansen 
Since this is consumable by end users, can we make this a `string` rather than a `obj list`.

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -597,16 +599,18 @@ let private executeSubscription (resultSet: (string * ExecutionInfo) []) (ctx: E
...
     let fieldCtx =
         { ExecutionInfo = info
           Context = ctx
           ReturnType = returnType
           ParentType = objdef
           Schema = ctx.Schema
           Args = args
-          Variables = ctx.Variables }
+          Variables = ctx.Variables
+          Path = fieldPath }
```
jberzy: I think fieldPath elements are in reverse order, so likely need to reverse before assigning.

jberzy:
@mickhansen Thanks for the PR! Just a couple of comments.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
mickhansen: @jberzy The error type expects an `obj list`: https://github.com/fsprojects/FSharp.Data.GraphQL/blob/3cd0940df3150be671c284d205f9071a5d5950ed/src/FSharp.Data.GraphQL.Server/Execution.fs#L16 which was my primary intended target for this PR.

src/FSharp.Data.GraphQL.Server/Execution.fs
mickhansen: Right! Fixed now.

jberzy:
@mickhansen thanks for the updates to the PR. What do you think about taking a try at solving the underlying problem with the error handling? I believe there are a few places where we only rely on exceptions to signal that the resolver has errors associated, but we should be inspecting the `ResolveFieldContext`'s Error list as well. e.g.:

https://github.com/fsprojects/FSharp.Data.GraphQL/blob/dev/src/FSharp.Data.GraphQL.Server/Execution.fs#L584

Thanks

mickhansen:
@jberzy It would be great to work on better error handling in general (especially for custom exceptions) - but this PR would be a good step to letting users add workarounds in the meantime.

In the project i'm working on we're using custom exceptions to signal that our monitoring system should ignore them (for instance ACL based errors should not trigger a pager duty signal), this information got lost if we simply did `failwith` but `ctx.AddError` lets us get around this, however we'd still like to add the errors basic to the result set in the end, which is where `Path´ comes in.

In general I think adding `Path` is a solid idea as it would also help any kind of resolve hook that needs to do diagnostics (like say ApplicationInsights resolver durations).

Another thing I'd like to look at is error handling from scalar coercion, specifically returning `None` to a non null field results in an unexpected state (the value ends up being null but it should probably throw with an error).

Adding an optional railway oriented approach to field resolvers is also on my potential "wish list" at the moment.

mickhansen:
I don't believe `ResolveFieldContext` currently has an error list, `AddError` adds to the overall `ExecutionContext`, however it could certainly do both. I'd be happy to look at that after this PR :)

mickhansen:
@jberzy Let me know if there's anything you need from me to get this merged.

jberzy:
The failure looks unrelated. Thanks for the contribution.

PR #346, opened 2021-08-31
fix-nullable-enum-input

mickhansen:
`Enum.IsDefined` will raise an exception if given a `null` value, currently making nullable enum inputs unusable.


mickhansen:
@jberzy keep-alive

PR #349, opened 2021-09-21
Fix recursive definition of InputObject fields

vanhoeij:
Using recursive definition of fields in the `InputObject` resulted in a null reference exception. The exception was triggered because the `ExecuteInput` on a field definition was not initialized. It was caused by not caching the field definitions correctly, resulting in uninitialized field definitions to be used.

In this PR, the function that is used to retrieve the definitions is wrapped inside a lazy, which makes sure that always the same field definitions are returned.

ps. it was required to update the dependencies because `MSBuild.StructuredLogger` was causing an issue when restoring the client project.

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
+++ b/src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
@@ -1528,15 +1528,15 @@ and InputObjectDefinition<'Val> =
       Name : string
       /// Optional input object description.
       Description : string option
-      /// Function used to define field inputs. It must be lazy
-      /// in order to support self-referencing types.
-      FieldsFn : unit -> InputFieldDef [] }
+      /// Lazy resolver for the input object fields. It must be lazy in
+      /// order to allow self-recursive type references.
+      FieldsFn : Lazy<InputFieldDef[]> }
     interface InputDef
 
     interface InputObjectDef with
         member x.Name = x.Name
         member x.Description = x.Description
-        member x.Fields = x.FieldsFn()
+        member x.Fields = x.FieldsFn.Force()
 
     interface TypeDef<'Val>
     interface InputDef<'Val>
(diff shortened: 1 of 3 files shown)


jberzy (changes requested):
Very Nice. Can we just remove that build.fsx.lock changes and I think this is good to go.

xperiandri:
@mickhansen, @nikhedonia any comments about this one?

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -1528,15 +1528,15 @@ and InputObjectDefinition<'Val> =
       Name : string
       /// Optional input object description.
       Description : string option
-      /// Function used to define field inputs. It must be lazy
-      /// in order to support self-referencing types.
-      FieldsFn : unit -> InputFieldDef [] }
+      /// Lazy resolver for the input object fields. It must be lazy in
+      /// order to allow self-recursive type references.
+      FieldsFn : Lazy<InputFieldDef[]> }
```
xperiandri:
```suggestion
      Fields : Lazy<InputFieldDef[]> }
```
As we release v2 I think we can rename it

xperiandri (changes requested):
@vanhoeij will you rebase this on top of the latest changes?

xperiandri:
@vanhoeij

xperiandri:
@vanhoeij

xperiandri:
@njlr what do you think?

njlr:
> @njlr what do you think?

This change makes sense to me. 

Good to fix the typos also :+1:

xperiandri:
@vanhoeij will you rebase?
Or give me access to your repo

xperiandri (approved):

xperiandri:
@vanhoeij thank you very much for your contribution!

xperiandri:
@vanhoeij 
```
Error Message:
   expected { fieldWithNestedInputObject: "{"na":{"a":null,"b":null,"c":"c","d":null},"nb":"b"}" }
    but got { fieldWithNestedInputObject: "{"na":{"a":null,"b":null,"c":"c","d":null,"e":null},"nb":"b"}" }
```

xperiandri:
Is the fix as simple as adding `,"e":null`?

xperiandri (approved):

PR #350, opened 2021-09-26
 * First pass at custom coerce for input objects

njlr:


src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -153,17 +158,29 @@ let rec private coerceVariableValue isNullable typedef (vardef: VarDef) (input:
...
+            coerceInput v |> Option.toObj
+        | _ -> failwithf "Expected an input Value but got %A" input
+    | None ->
+        match input with
+        | :? Map<string, obj> as map ->
+            let mapping =
+                objdef.Fields
+                |> Array.map (fun field ->
+                    let valueFound = Map.tryFind field.Name map |> Option.toObj
+                    let coercedValue = coerceVariableValue false field.TypeDef vardef valueFound (errMsg + (sprintf "in field '%s': " field.Name))
```
xperiandri, line 175:
```suggestion
                |> Seq.map (fun field ->
                    let valueFound = map |> Map.tryFind field.Name |> Option.toObj
                    let coercedValue = coerceVariableValue false field.TypeDef vardef valueFound (errMsg + (sprintf "in field '%s': " field.Name))
```
Maybe it does not worth to materialize it to an array

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -1530,13 +1532,22 @@ and InputObjectDefinition<'Val> =
       Description : string option
       /// Function used to define field inputs. It must be lazy
       /// in order to support self-referencing types.
-      FieldsFn : unit -> InputFieldDef [] }
+      FieldsFn : unit -> InputFieldDef []
+      /// Optional function to customize default coercion behaviour
+      CoerceInput : (Value -> 'Val option) option }
```
xperiandri, line 1537: Have you considered using `Result<,>`?
Or @jberzy are there other places to move to `Result` so it is better to have a new PR for that?

xperiandri (commented):
I like what you have done. This is useful.
Also, it would be more useful to be able to return an error and pass that error to the response with error to GraphQL field mapping.

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -38,31 +38,36 @@ let rec internal compileByType (errMsg: string) (inputDef: InputDef): ExecuteInp
     | Scalar scalardef ->
         variableOrElse (scalardef.CoerceInput >> Option.toObj)
```
xperiandri, line 39: @jberzy can we switch this to `Result<,>` and pass an error to the response?

xperiandri:
There is an issue I have with current scalar implementation
``` F#
    interface ScalarDef with
        member x.Name = x.Name
        member x.Description = x.Description
        member x.CoerceInput input = x.CoerceInput input |> Option.map box
        member x.CoerceValue value = (x.CoerceValue value) |> Option.map box
```

This piece is very contraintuitive
``` F#
        member x.CoerceInput input = x.CoerceInput input |> Option.map box
        member x.CoerceValue value = (x.CoerceValue value) |> Option.map box
```
@jberzy at least XML comments with clear description required.
Can we get rid of `CoerceValue` at all?

And `CoerceInput` here may be optional. Because if I preprocess variables with JSON deserializer I don't need any coercion at all.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
njlr: I would also prefer result, although the current design is based around options (see coerce input methods)

njlr:
@xperiandri I have fleshed out this idea further: https://github.com/fsprojects/FSharp.Data.GraphQL/pull/351

PR #351, opened 2021-10-06
Decoders

njlr:
This PR introduces "decoders" to input types in the schema. 

A decoder is like a `coerceInput` function, but it returns a structured `Error` rather than an `option`:

```fsharp
type DecodeResult<'t> = Result<'t, DecodeError>

type Decoder<'t> = Value -> DecodeResult<'t>
```

A helper library of decoders is provided for common types like `string`, `int`, etc. There is also a utility for automatic decoder generation using reflection. 

All input types (not just scalars) now have a decoder attached to them. 

The user can provide a decoder (recommended) or they can allow a default one to be generated automatically. This allows the user to interpret the GraphQL AST _with the context of the type_ being decoded. 

There are currently some scrappy comments etc. If we like the idea I can clean it up before merging.


src/FSharp.Data.GraphQL.Server/Relay/Node.fs
```diff
@@ -45,15 +45,15 @@ module GlobalId =
         static member GlobalIdField (typeName: string, resolve: (ResolveFieldContext -> 'In -> string)) =
             Define.Field(
                 name = "id",
-                typedef = ID,
+                typedef = ID<string>,
```
xperiandri, line 48: As far as I understand the intent of GraphQL designers, ID was supposed to be a GUID

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -8,8 +8,54 @@ open System
...
+    [
+        for p in error.Path do
+            match p with
+            | AtIndex index -> sprintf "list element %i" index
+            | InProperty name -> sprintf "in field '%s'" name
+            | InInputObject typeName -> sprintf "in input object '%s'" typeName
+
+        match error.Reason with
+        | DecodeErrorReason.TypeMismatch (expected, actual) ->
+            sprintf "expected value of type %s but got %s" expected actual
```
xperiandri, line 26: Is this phrase correct?

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -76,18 +136,30 @@ let rec internal compileByType (errMsg: string) (inputDef: InputDef): ExecuteInp
...
+            match value with
+            | Variable name ->
+                match Map.tryFind name vars with
+                | Some o -> o
+                | None -> null
+            | value ->
+                match nullableDef.BoxedDecoder value with
+                | Ok t -> t
+                | Error error ->
+                    raise(GraphQLException (errMsg + (string error))))
```
xperiandri, line 153: What is `GraphQLException`? Is it something that existed in this library?

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -165,17 +237,49 @@ and private coerceVariableInputObject (objdef) (vardef: VarDef) (input: obj) err
...
+            (match Map.tryFind vname inputs with
+            | Some value when decoderAllowsNull || value <> NullValue -> Ok value
+            | _ ->
+                match varDef.DefaultValue with
+                | Some defaultValue -> Ok defaultValue
+                | None ->
+                    match varDef.TypeDef with
+                    | Nullable _ -> Ok NullValue
+                    |_ ->
+                        let message = sprintf "Variable '$%s': expected value of type %s, but no value was found" vname (varDef.TypeDef.ToString())
```
xperiandri, line 256: Maybe it is better to use string interpolation?

src/FSharp.Data.GraphQL.Shared/Decoding.fs
```diff
@@ -0,0 +1,618 @@
...
+    {
+      Reason = DecodeErrorReason.Failure reason
+      Path = []
+    }
+
+  let nest (pathSegment : DecodePathSegment) (error : DecodeError) =
+    {
+      error with
+        Path = pathSegment :: error.Path
+    }
```
xperiandri: Maybe in a single line?
```suggestion
    { error with Path = pathSegment :: error.Path }
```

src/FSharp.Data.GraphQL.Shared/Decoding.fs
```diff
@@ -0,0 +1,618 @@
...
+
+[<AutoOpen>]
+module ComputationExpression =
+
+  type DecoderBuilder() =
+    member this.Bind(m, f) =
+      Decoder.bind f m
+
+    member this.Return(x) =
+      Decoder.succeed x
```
xperiandri: I would format this more compactly
```suggestion
  type DecoderBuilder() =
    member this.Bind(m, f) = Decoder.bind f m
    member this.Return(x) = Decoder.succeed x
```

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -2782,11 +3101,17 @@ module SchemaDefinitions =
...
-            { Name = name
+        static member InputObject(name : string, fieldsFn : unit -> InputFieldDef list, decoder : Decoder<'Out>, ?description : string) : InputObjectDefinition<'Out> =
+            {
+              Name = name
               FieldsFn = fun () -> fieldsFn() |> List.toArray
-              Description = description }
+              Description = description
+              Decoder =
+                decoder
+                |> Decoder.mapError (DecodeError.nest (InInputObject name))
```
xperiandri: I usually write such constructs with a pipe to reduce braces amont
```suggestion
                |> Decoder.mapError (DecodeError.nest <| InInputObject name)
```

xperiandri (commented):
Looks interesting.
But I have not noticed a practical usage sample in tests.

src/FSharp.Data.GraphQL.Server/Values.fs
njlr: Yes they were chosen to give minimal changes in the tests that check error messages

src/FSharp.Data.GraphQL.Server/Values.fs
njlr: Yes that was already there

src/FSharp.Data.GraphQL.Server/Values.fs
njlr: Version of F# / .NET that the library targets doesn't support them, although I would prefer them too!

xperiandri: @jberzy what about updating `FSharp.Core` to 5.0 at least or better 6.0?

src/FSharp.Data.GraphQL.Server/Relay/Node.fs
njlr: From [this](https://graphql.org/learn/schema/):

> ID: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an ID signifies that it is not intended to be human‐readable.

So I think a string is the closest .NET type.

njlr:
> Looks interesting. But I have not noticed a practical usage sample in tests.

Good point - I can update one of the samples perhaps

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri, line 26:
```suggestion
            sprintf "expected value of type '%s' but got '%s'" expected actual
```

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri, line 256:
```suggestion
                        let message = sprintf "Variable '$%s': expected value of type '%s', but no value was found" vname (varDef.TypeDef.ToString())
```

xperiandri (changes requested):

PR #357, opened 2022-02-23
support using DUs as Enum variables

mickhansen:


sgryt:
@mickhansen Looks good to me.

tests/FSharp.Data.GraphQL.Tests/VariablesTests.fs
```diff
@@ -557,3 +557,42 @@ let ``Execute handles nullable enum input as variable`` () =
...
+      data.["data"] |> equals (upcast expected)
+    | _ -> fail "Expected Direct GQResponse"
+
+[<Fact>]
+let ``Execute handles Some union enum input as variable`` () =
+    let ast = parse """query fieldWithNullableEnumValue($enumVar: EnumTestType) {
+        fieldWithNullableEnumInput(input: $enumVar)
+      }"""
+    let actual = sync <| Executor(schema).AsyncExecute(ast, variables = Map.ofList ["enumVar", Some Bar :> obj ])
+    let expected = NameValueLookup.ofList [ "fieldWithNullableEnumInput", upcast "{\"case\":\"Bar\"}" ]
```
jberzy, line 579: Since DUs are represented differently than Enums, and that only DUs that don't contain parameters are valid in this case, maybe we should add a JsonSerializer that overrides the default Json.NET encoding and strips the outer "case" portion i.e. the "Tag" becomes the json string result making the output identical to the Enum handler.

build.fsx
```diff
@@ -103,18 +103,23 @@ Target.create "CleanDocs" (fun _ ->
 Target.create "Restore" (fun _ ->
     !! "src/**/*.??proj"
     -- "src/**/*.shproj"
-    |> Seq.iter (DotNet.restore id))
+    |> Seq.iter (fun pattern ->
+        DotNet.restore (fun options ->
+            { options with MSBuildParams = { options.MSBuildParams with DisableInternalBinLog = true } }
```
jberzy, line 108: Does that solve a build issue?

jberzy (changes requested):

build.fsx
mickhansen: @jberzy Yes it solved the OS X build issue we've been having on GitHub (and also solves a local build issue of mine)

tests/FSharp.Data.GraphQL.Tests/VariablesTests.fs
mickhansen: @jberzy We could but I don't think it has any meaningful implications for production code. The resolver for this test simply stringifies whatever it's given and that's why we end up with `case:`, I don't think people would write resolvers like that in real code, and if they did they can control how it's stringified.

mickhansen: What I mean is that this output is the direct result of using `stringifyInput` further up in the test code, it's not being produced by the library itself.

jberzy: Ok, that's fine. We can have a separate PR address adding custom Json Converters. Maybe after we merge the support for System.Text.Json.

jberzy:
LGTM. Thanks!

PR #366, opened 2022-05-17
bugfix/issue-365-enum-encoding

njlr:
* Enums now encoded using name
 * Adds test

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Execution.fs
+++ b/src/FSharp.Data.GraphQL.Server/Execution.fs
@@ -322,7 +322,7 @@ let rec private direct (returnDef : OutputDef) (ctx : ResolveFieldContext) (path
         | Some v' -> resolved name v'
         | None -> raiseError <| coercionError value scalarDef.Name path ctx
     | Enum enumDef ->
-        let enumCase = enumDef.Options |> Array.tryPick(fun case -> if case.Value.Equals(value) then coerceStringValue value else None)
+        let enumCase = enumDef.Options |> Array.tryPick(fun case -> if case.Value.Equals(value) then Some case.Name else None)
         match enumCase with
         | Some v' -> resolved name (v' :> obj)
         | None -> raiseError <| coercionError value enumDef.Name path  ctx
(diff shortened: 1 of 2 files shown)


xperiandri:
Look great. Is it ready for merge?

xperiandri:
@nikhedonia, @mickhansen any comments or suggestions?

njlr:
> Look great. Is it ready for merge?

I think so!

Main thing to check is if the new behavior is exactly what we want as this will be a breaking change in _some_ cases. 

1. Behavior is the same:

```fsharp
type private Fruit =
  | Apple
  | Banana
  | Cherry

let private enumType =
  Define.Enum(
    "Fruit",
    [
      Define.EnumValue("Apple", Apple)
      Define.EnumValue("Banana", Banana)
      Define.EnumValue("Cherry", Cherry)
    ]
  )
```

2. Behavior is the same:

```fsharp
type private Fruit =
  | Apple
  | Banana
  | Cherry
  with
    override this.ToString() = 
      match this with
      | Apple -> "APPLE"
      | Banana -> "BANANA"
      | Cherry -> "CHERRY"

let private enumType =
  Define.Enum(
    "Fruit",
    [
      Define.EnumValue("APPLE", Apple)
      Define.EnumValue("BANANA", Banana)
      Define.EnumValue("CHERRY", Cherry)
    ]
  )
```

3. Behavior is different:

```fsharp
type private Fruit =
  | Apple
  | Banana
  | Cherry

let private enumType =
  Define.Enum(
    "Fruit",
    [
      Define.EnumValue("APPLE", Apple)
      Define.EnumValue("BANANA", Banana)
      Define.EnumValue("CHERRY", Cherry)
    ]
  )
```

xperiandri:
While working with GraphQL and this library in particular I found that I must validate a scheme with the original JavaScript library via Node.js and perform integration tests from a generated client down to a database gathering all the available fields.
Only that ensures that everything works as expected.

mickhansen:
Great! `Define.EnumValue("APPLE", Apple)` definitely threw me for a loop when I first started implementing.

Only question is how we handle breaking changes, where are we in terms of publishing a new nuget package? This might require a major version bump.

Not even sure if we need to consider this a breaking change, more of a bug fix since the `Define.EnumValue("APPLE", Apple)` case never worked properly IIRC.

nikhedonia:
I like this and happy to merge

PR #377, opened 2022-08-22
fix(376): rec func "go" was misintrepreting param "ms" (middlewares)

valbers:
Issue found while trying to run the Star Wars API sample.
 Somehow, in the pattern matching branch "None -> go ctx ms", "go" was being
reevaluated to be a function whose ctx parameter is of type unit and the
'res type of the containing function runMiddlewares was being
reevaluated to be IExecutorMiddleware (something like a parameterless ExecutorMiddleware
fetcher), although ctx was clearly initially set to be a
SchemaCompileContext and 'res a "unit" because of the type of the
"compileSchema" function passed as parameter in a later function call.
 However, ms is always supposed to be a list of middlewares yet it was
being defined as a generic list, more specifically: the type definition was implicit. By adding
an explicit hint to the compiler in the code that "m" is an IExecutorMiddleware (and therefore "ms" a list thereof),
the compiler no longer had a misinterpretation issue with the recursive "go" function
and the Star Wars API sample now starts successfully.

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Executor.fs
+++ b/src/FSharp.Data.GraphQL.Server/Executor.fs
@@ -74,9 +74,10 @@ type Executor<'Root>(schema: ISchema<'Root>, middlewares : IExecutorMiddleware s
                            (initialCtx : 'ctx)
                            (onComplete : 'ctx -> 'res)
                            : 'res =
-        let rec go ctx = function
+        let rec go ctx (middlewares: IExecutorMiddleware list) =
+            match middlewares with
             | [] -> onComplete ctx
-            | m : IExecutorMiddleware :: ms ->
+            | m :: ms ->
                 match (phaseSel m) with
                 | Some f -> f ctx (fun ctx' -> go ctx' ms)
                 | None -> go ctx ms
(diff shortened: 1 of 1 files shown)


src/FSharp.Data.GraphQL.Server/Executor.fs
```diff
@@ -76,7 +76,7 @@ type Executor<'Root>(schema: ISchema<'Root>, middlewares : IExecutorMiddleware s
                            : 'res =
         let rec go ctx = function
             | [] -> onComplete ctx
-            | m :: ms ->
+            | m : IExecutorMiddleware :: ms ->
```
xperiandri:
```suggestion
            | (m : IExecutorMiddleware) :: ms ->
```
for readability

xperiandri:
@mickhansen, @nikhedonia any suggestions?

src/FSharp.Data.GraphQL.Server/Executor.fs
xperiandri:
```suggestion
        let rec go ctx (middlewares : IExecutorMiddleware list) =
            match middleware with
            | [] -> onComplete ctx
            | m :: ms ->
```
or maybe like this

xperiandri (approved):

valbers:
@xperiandri thank you for approving the pull request so quickly. Is the paket problem with the ubuntu CI pipeline known and should be ignored or can you take a look into that (maybe rerun the build job)?

xperiandri:
I haven't set up that. I guess we had that issue previously.
Actually I want to get rid of pocket and move to packages.props

xperiandri:
Thank you very much for the contribution!

valbers:
I'm glad I could contribute in some way. Is it possible to link this pull request to the original issue and close the issue, so that other people know that this has already been taken care of?

Issue #328, opened 2021-04-03
Remove JsonValue.fs from client and use System.Text.Json


xperiandri:
### Description

The JsonValue.fs deserializer deserializes `0` into `int` while it must deserialize it into the type requested by the schema. Which is `float` in my case.
We can simply use System.Text.Json package version 5 as it is done in https://github.com/fsprojects/SwaggerProvider/tree/net5 with https://github.com/Tarmil/FSharp.SystemTextJson/

### Repro steps

1. Define `Float` field in GraphQL and return `0.0` from it
2. Use System.Text.Json as a serializer
3. You will get `int option` 

### Expected behavior

Value is deserialized into the type defined by the schema

### Actual behavior

Value is deserialized into the type chosen by JsonValue.fs logic

### Related information 

1.0.7

njlr:
There are a few things that the homegrown JSON (and the more widely used `Thoth.Json.Net`) offers:

 1. Fine-grained control over encoding and decoding without using reflection, annotations or writing "DTO" types:

```fsharp
type Money = 
  {
    Dollars : int
    Cents : int
  }

// JSON representation is:
//     "123.45"

let decode = 
  decoder {
    let! (x : string) = Decode.string

      match x.Split([| '.' |]) with
      | [| dollars; cents |] ->
        let! dollars = 
          match Int32.TryParse(dollars) with
          | true, i -> Decode.succeed i
          | _ -> Decode.fail "Invalid dollars"

        let! cents = 
          match Int32.TryParse(cents) with
          | true, i -> Decode.succeed i
          | _ -> Decode.fail "Invalid cents"

        return { Dollars = dollars; Cents = cents }
      | _ ->
        return! Decode.fail "Expected <dollars>.<cents>"
  }
```

2. Ability to visit a JSON structure in a type-safe manner using `match`

```fsharp
let json = Decode.unsafeFromString "{ \"foo\": 123 }"

match json with
| JsonValue.Object m ->
   match Map.tryFind "foo" m with
   | Some (JsonValue.Integer x) -> x
   | _ -> 0
| _ -> 0
```

3. Ability to compare JSON values as values:

```fsharp
let a = Encode.object [ "foo", Encode.int 123 ]
let b = Encode.object [ "foo", Encode.int 123 ]

Assert(a = b)
```

4. Support for Fable - we should at some point bring back support for Fable in the client type-provider

Does `System.Text.Json` offer these?

[...]

PR #400, opened 2022-10-24
Specify signature file for JsonValue

wallymathieu:
In order to make it easier to swap out the implementation, let us make it more obvious what is used from JsonValue
See #328


xperiandri:
Let me set up GraphQL on my current project with that latest changes from @njlr and after that I will have some view on changes I like to see in JSON interop.
I'll ping you here with a branch you can have a look and propose yours.

wallymathieu:
I think I have forgotten why I did this, let us close this so that it does not disrupt

PR #403, opened 2022-11-27
Added NuGet dependencies between packages

Juriyx:


MaxymVlasov (commented):
@webknjaz please take a look when you will have a free minute. Maybe you could suggest some improvements.
Tnx

xperiandri (approved):

xperiandri (approved):

.github/workflows/publish_ci.yml
```diff
@@ -8,27 +8,43 @@ on:
 jobs:
   build:
 
+    env:
+      GITHUB_TOKEN: '${{secrets.GITHUB_TOKEN}}'
```
webknjaz, line 12: Putting secrets into globally available env vars is not the best idea ever.

.github/workflows/publish_release.yml
```diff
@@ -8,41 +8,40 @@ on:
...
-          VERSION_REGEX: <Version>(.*)<\/Version>
-          NUGET_KEY: ${{secrets.NUGET_SECRET}}
-      - name: publish FSharp.Data.GraphQL.Server.Middleware on version change
-        uses: rohith/publish-nuget@v2
-        with:
-          PROJECT_FILE_PATH: src/FSharp.Data.GraphQL.Server.Middleware/FSharp.Data.GraphQL.Server.Middleware.fsproj
-          VERSION_FILE_PATH: Directory.Build.targets
-          VERSION_REGEX: <Version>(.*)<\/Version>
-          NUGET_KEY: ${{secrets.NUGET_SECRET}}
+    - uses: actions/checkout@v2
```
webknjaz, line 17: This is a deprecated action version that causes warnings. It should be bumped.

.github/workflows/publish_ci.yml
```diff
@@ -8,27 +8,43 @@ on:
...
+
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
-    - name: Pack FSharp.Data.GraphQL.Client project
+
+    - name: Add GitHub Packges NuGet source
+      run: dotnet nuget add source --username USERNAME --password ${{ secrets.GITHUB_TOKEN }} --store-password-in-clear-text --name github "https://nuget.pkg.github.com/fsprojects/index.json"
+
+    - name: Set version
+      run: sed -i "s|<Version>\(.*\)</Version>|<Version>\1-ci-$GITHUB_RUN_ID</Version>|" Directory.Build.targets
```
webknjaz, line 26: It might be a good idea to generate version using something like dunamai.

.github/workflows/publish_release.yml
```diff
@@ -8,41 +8,40 @@ on:
...
+        cd src/FSharp.Data.GraphQL.Server
+        dotnet pack --nologo --configuration Release /p:IsNuget=true -o ../../nuget
+        dotnet nuget push ../../nuget/*Server*.nupkg -k $KEY
+
+    - name: Pack FSharp.Data.GraphQL.Server.Middleware project and push to NuGet
+      if: startsWith(github.ref, 'refs/tags/release')
+      run: |
+        cd src/FSharp.Data.GraphQL.Server.Middleware
+        dotnet pack --nologo --configuration Release /p:IsNuget=true -o ../../nuget
+        dotnet nuget push ../../nuget/*Server.Middleware*.nupkg -k $KEY
```
webknjaz, line 47: It's typically good to build the project in a separate job, save it as an artifact, test that artifact, and publish the same tested artifact only after that. All in the same workflow. To reduce duplication, reusable workflows can be used, they good enough these days.

xperiandri: Could you show an example of that?

.github/workflows/publish_ci.yml
xperiandri: What is a better idea? Just access like this `${{secrets.GITHUB_TOKEN}}`?

webknjaz: Yes, or only set the env var for specific steps if really necessary.

xperiandri: Could you show how to do that?

.github/workflows/publish_release.yml
webknjaz: I only have examples from the Python ecosystem, they will probably not help you. But it's easy. Use the `needs` attribute to chain stages and use `actions/upload-artifact` in the build job and `actions/download-artifact` in the test and publish ones.
I don't know anything about the dotnet ecosystem to suggest how to actually invoke testing — you'll need to figure it out yourself.

.github/workflows/publish_ci.yml
webknjaz: Setting an env var? Same as here but in the step definitions needing an env var. Although, I saw they already access the token through the secrets context so I don't know why you additionally put it into an env var for everyone to read.

webknjaz: Looking at this more, I realize that the workflow-global env may not even have access to the `secrets` context so it's possible that you're setting the var to an empty string already.

webknjaz (commented):
Two more possibilities to tighten security:
1) use the `permissions` attribute on the job level to reset any unnecessary privileges and only keep what the token needs
2) integrate https://app.stepsecurity.io to only allow know outgoing traffic

.github/workflows/publish_release.yml
Juriyx: @webknjaz Hello. Could you, please, send me examples from the Python ecosystem?

PR #427, opened 2023-02-24
`ID` type deserialization fix with field tests

xperiandri:
@ivelten I think testing should happen from here.
Also, we can change some `id` fields on StarWars API to cover different scenarios

### DIFF
--- a/src/FSharp.Data.GraphQL.Client/BaseTypes.fs
+++ b/src/FSharp.Data.GraphQL.Client/BaseTypes.fs
@@ -348,7 +348,7 @@ module internal JsonValueHelper =
                             | Some "Date" ->
                                 match DateTime.TryParse(s, CultureInfo.InvariantCulture, DateTimeStyles.None) with
                                 | (true, d) -> box d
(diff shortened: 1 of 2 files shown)


xperiandri:
Probably we need to modify some tests too

src/FSharp.Data.GraphQL.Client/BaseTypes.fs
```diff
@@ -357,14 +357,16 @@ module internal JsonValueHelper =
                     | None -> failwith "Item type is a non null type, but no underlying type exists on the schema definition of the type."
                 | TypeKind.SCALAR ->
                     match schemaField.SchemaTypeRef.Name with
+                    | Some "String" | Some "ID" ->
+                        s |> makeSomeIfNeeded
```
valbers, line 361: did this really fix the issue? This surprises me, since there is already an arm covering both cases (the default one) and doing the same.

valbers (changes requested):
Please respond my comment.

valbers:
I'll believe in you and assume I'm missing something. Let's close this PR.

valbers (approved):

PR #436, opened 2023-08-07
Input object creation with fields coercion

xperiandri:


xperiandri:
@valbers have a look if you have time

.editorconfig
```diff
@@ -187,17 +182,29 @@ fsharp_max_function_binding_width=80
...
+#                 {
+#                     From: float
+#                     To: float
+#                 }
+# stroustrup: type Range = {
+#                 From: float
+#                 To: float
+#             }
+# default cramped
+fsharp_multiline_bracket_style=stroustrup
```
valbers, line 199: Isn't it supposed to be? :thinking:
```suggestion
fsharp_multiline_bracket_style=cramped
```

valbers:
While I was taking a look, I pulled the changes to a local working copy and tried to build the solution (using `./build.sh` on mac os), but some projects were failing seemingly because of conflicting `FSharp.Core` versions between those referenced by nuget packages (7.0.0) and the one referenced by us directly (6.0.3). I tried to fix that and ended up having to downgrade `FsToolkit.ErrorHandling` to 2.13. My suggestion would be to keep the older version if possible and user newer versions with a new `FSharp.Data.GraphQL` major release (which doesn't need to happen much later IMHO).
After that, there were other errors:
- Incomplete pattern matches on `InputParameterValue` (Various lines in `TypeSystem.fs`; case `Variable (_)` was missing);
- `SchemaDefinitions` GraphQL type definitions were renamed (the suffix "Type" was added to their names, like in `String` ->  `StringType`) but not updated in `src/FSharp.Data.GraphQL.Server.Relay/Connections.fs`.

Directory.Build.targets
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <LangVersion>6.0</LangVersion>
     <FSharpCoreVersion>6.0.*</FSharpCoreVersion>
+    <AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
```
valbers, line 7: Love it. That's when OSS pays off: you're always learning something new :blush:.

xperiandri:
@valbers, I fixed the second one

xperiandri:
@njlr could you review too?

valbers:
@xperiandri great! Now what about the `FsToolkit.ErrorHandling` issue?

xperiandri:
I’m working on that.
Is it only for Type Provider?

valbers:
> I’m working on that. Is it only for Type Provider?

I'm not sure what you mean. Anyway, it's about the errors one gets when running `./build.sh`.

xperiandri:
Added some fixes. But integration tests are fixed in #436

xperiandri:
@valbers @njlr @ivelten @mickhansen this PR is ready for review

samples/star-wars-api/Ast.fs
```diff
@@ -0,0 +1,38 @@
...
+
+let findOperationByName operationName (astDoc: Ast.Document)  =
+    match astDoc.Definitions |> List.choose getOperation, operationName with
+    | [ def ], _ -> Some def
+    | defs, name -> defs |> List.tryFind (fun def -> def.Name = name)
+
+let containsFieldsBeyond
+    (allowedFields: ImmutableHashSet<_>)
+    (onSuccess: Ast.Field -> unit)
+    (onFail: unit -> unit)
```
valbers: I'm not sure if `onSuccess` and  `onFail` convey the real meaning of these functions.
How about `onEachInspectedField` and `onEachInspectedNonField` respectively?

samples/star-wars-api/JSON.fs
```diff
@@ -10,23 +10,29 @@ let [<Literal>] UnionTag = "kind"
...
+        UnionTag,
+        allowOverride = true)
 
 let configureDefaultSerializerOptions = configureSerializerOptions defaultJsonFSharpOptions
 
 let getSerializerOptions (additionalConverters: JsonConverter seq) =
-    JsonSerializerOptions () |> configureDefaultSerializerOptions additionalConverters
+    let options = JsonSerializerOptions ()
+    options |> configureDefaultSerializerOptions additionalConverters
+    options
```
valbers, line 36: Is this a debugging left-over? I'd have nothing against the previous form.

src/FSharp.Data.GraphQL.Shared/Prolog.fs
```diff
@@ -22,15 +23,20 @@ type MalformedQueryException(msg) =
...
         if isNull value then None
         else
             let t = value.GetType()
-            let p = t.GetProperty("Value")
-            if t.IsGenericType && t.GetGenericTypeDefinition() = optionDef then
+            if t.Name = "FSharpOption`1" then
+                let p = t.GetProperty("Value")
                 Some (p.GetValue(value, [||]))
+            elif t.Name = "FSharpValueOption`1" then
+                if value = Activator.CreateInstance t then None
```
valbers, line 36: would maybe `if value = upcast ValueNone then None` also have worked?

xperiandri:
@valbers @njlr @ivelten @mickhansen any feedback?

tests/FSharp.Data.GraphQL.Tests/CoercionTests.fs
```diff
@@ -14,10 +14,10 @@ open FSharp.Data.GraphQL.Types.Patterns
 
 let private testCoercion graphQLType (expected: 't) actual =
     let (Scalar scalar) = graphQLType
-    let result = (scalar.CoerceInput actual) |> Option.map (fun x -> downcast x)
+    let result = (scalar.CoerceInput actual) |> Result.map (fun x -> downcast x)
     match result with
-    | Some x -> equals expected x
-    | None -> raise (Exception(sprintf "Expected %A to be able to be coerced to %A" actual expected))
+    | Ok x -> equals expected x
+    | Error _ -> raise (Exception(sprintf "Expected %A to be able to be coerced to %A" actual expected))
```
valbers, line 20: Maybe it'd be helpful to also show the error in the exception right away?
I mean:
```fsharp
raise (Exception(sprintf "Expected %A to be able to be coerced to %A, but got %A" actual expected result))
```

src/FSharp.Data.GraphQL.Server.Middleware/MiddlewareDefinitions.fs
```diff
@@ -80,29 +80,36 @@ type internal ObjectListFilterMiddleware<'ObjectType, 'ListType>(reportToMetadat
...
+                        | "filter" -> ObjectListFilter.CoerceInput (InlineConstant x.Value)
+                        | _ -> Ok NoFilter)
+                match filterResults |> splitSeqErrorsList with
+                | Error errs -> Error errs
+                | Ok filters ->
+                    filters
+                    |> removeNoFilter
+                    |> Seq.map (fun x -> field.Ast.AliasOrName, x)
+                    |> Seq.toList
+                    |> Ok
```
valbers, line 97: How about the following instead (i.e. using `Result.map`)?
```fsharp
let filterResults =
   field.Ast.Arguments
   |> Seq.map (fun x ->
         match x.Name with
         | "filter" -> ObjectListFilter.CoerceInput (InlineConstant x.Value)
         | _ -> Ok NoFilter)
   |> splitSeqErrorsList
   |> Result.map // <--- HERE
       (removeNoFilter
         >> Seq.map (fun x -> field.Ast.AliasOrName, x)
         >> Seq.toList)
```

src/FSharp.Data.GraphQL.Server.Middleware/MiddlewareDefinitions.fs
```diff
@@ -80,29 +80,36 @@ type internal ObjectListFilterMiddleware<'ObjectType, 'ListType>(reportToMetadat
...
+                        collectArgs acc fields
+                    | ResolveCollection field ->
+                        fieldArgs field
+                    | ResolveAbstraction typeFields ->
+                        let fields = typeFields |> Map.toList |> List.collect (fun (_, v) -> v)
+                        collectArgs acc fields
+                    | _ -> Ok acc
+                match accResult with
+                | Error errs -> Error errs
+                | Ok acc -> collectArgs acc xs
```
valbers, line 113: Again, what do you think about using `Result.map` instead?
```fsharp
  accResult
  |> Result.map (fun x -> collectArgs x xs)
```

src/FSharp.Data.GraphQL.Server.Middleware/ObjectListFilter.fs
```diff
@@ -17,6 +17,8 @@ type ObjectListFilter =
     | EndsWith of FieldFilter<string>
     | Contains of FieldFilter<string>
     | FilterField of FieldFilter<ObjectListFilter>
+    | NoFilter
```
valbers, line 20: What do you think about defining [neutral elements](https://en.wikipedia.org/wiki/Identity_element) always as the first case of discriminated unions?

src/FSharp.Data.GraphQL.Server.Middleware/SchemaDefinitions.fs
```diff
@@ -1,13 +1,20 @@
...
-open System
+open FSharp.Data.GraphQL.Errors
+
 
 /// Contains customized schema definitions for extensibility features.
 [<AutoOpen>]
 module SchemaDefinitions =
-    let rec private coerceObjectListFilterInput x =
+
+    let internal removeNoFilter = Seq.where (fun filter -> filter <> NoFilter)
```
valbers, line 14: Just an idea: instead of defining a new function just for that one could call `Seq.where` in a more compact way:
```fsharp
Seq.where ((<>) NoFilter)
```

src/FSharp.Data.GraphQL.Server.Middleware/SchemaDefinitions.fs
```diff
@@ -30,63 +38,82 @@ module SchemaDefinitions =
...
             | FloatValue v -> ComparableValue (v :> System.IComparable)
             | BooleanValue v -> ComparableValue (v :> System.IComparable)
             | StringValue v -> ComparableValue (v :> System.IComparable)
             | v -> Other v
+
         let buildAnd x =
             let rec build acc x =
                 match x with
                 | [] -> acc
                 | x :: xs ->
```
valbers, line 54: I know it's not part of your changes, but while we're at it: I find the shadowing of `x` unnecessarily confusing here.
What do you think about renaming the first `x` to `tail` (or `remainingTail`)? It would look like this:
```fsharp
let rec build acc tail =
   match tail with
   | [] -> acc
   | x :: xs ->
      match acc with
      | NoFilter -> build x xs
      | _ -> build (And (acc, x))) xs
```

src/FSharp.Data.GraphQL.Server.Middleware/SchemaDefinitions.fs
```diff
@@ -30,63 +38,82 @@ module SchemaDefinitions =
...
-            build None x
+                    | NoFilter -> build (x) xs
+                    | acc -> build ((And (acc, x))) xs
+            build NoFilter x
+
         let buildOr x =
             let rec build acc x =
                 match x with
                 | [] -> acc
                 | x :: xs ->
```
valbers, line 64: Here again `x` as the tail being shadowed by `x`, the first element in the tail, makes the code unnecessarily harder to read.

src/FSharp.Data.GraphQL.Server.Middleware/SchemaDefinitions.fs
```diff
@@ -30,63 +38,82 @@ module SchemaDefinitions =
...
+                    | acc -> build ((Or (acc, x))) xs
+            build NoFilter x
+
         let rec mapFilter (name : string, value : InputValue) =
-            let mapFilters fields = fields |> List.map coerceObjectListFilterInput |> List.choose id
+            let mapFilters fields =
+                let coerceResults = fields |> Seq.map coerceObjectListFilterInput |> splitSeqErrorsList
+                match coerceResults with
+                | Error errs -> Error errs
+                | Ok coerced -> coerced |> removeNoFilter |> Seq.toList |> Ok
```
valbers, line 75: `Result.map` FTW 😉 (like you did in lines 77 and 78).

valbers:
@xperiandri I'm taking a look at it. It's a lot of changes, so it will take some time but I'm on it. Can you see some of my comments yet? BTW I'm currently not a "requested reviewer" for this PR.

xperiandri:
I cannot add you unless you make a review yourself first or become a repo member (see invite)

xperiandri:
> @xperiandri great! Now what about the `FsToolkit.ErrorHandling` issue?

What was your issue? It works fine for me now

valbers:
> > @xperiandri great! Now what about the `FsToolkit.ErrorHandling` issue?
> 
> What was your issue? It works fine for me now

The issue was with the incompatibility with .NET 6.0, but now you updated the target framework to .NET 7.0 so it should be no problem now.

xperiandri:
But .NET 6.0 builds too. As I added multitargeting

tests/FSharp.Data.GraphQL.IntegrationTests.Server/Properties/launchSettings.json
```diff
@@ -21,7 +21,8 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
-      "applicationUrl": "http://localhost:49245/"
+      "applicationUrl": "http://localhost:8085/"
+      //"applicationUrl": "http://localhost:49245/"
```
valbers, line 25: Hmm... This is actually invalid JSON 🙂 . Do you mind removing it?

samples/star-wars-api/Schema.fs
```diff
@@ -1,5 +1,16 @@
 namespace FSharp.Data.GraphQL.Samples.StarWarsApi
 
+open System
+open Microsoft.AspNetCore.Http
+open Microsoft.Extensions.DependencyInjection
+
+type Root(ctx : HttpContext) =
```
valbers, line 7: Good job. It makes so much more sense to have `HttpContext` as part of the `Root` object.

samples/star-wars-api/Startup.fs
```diff
@@ -18,14 +18,29 @@ type Startup private () =
...
+            .Configure(Action<KestrelServerOptions>(fun x -> x.AllowSynchronousIO <- true))
+            .Configure(Action<IISServerOptions>(fun x -> x.AllowSynchronousIO <- true))
+            // Surprisingly minimal APIs use Microsoft.AspNetCore.Http.Json.JsonOptions
+            // Use if you want to return HTTP responses using minmal APIs IResult interface
+            .Configure<HttpClientJsonOptions>(
+                Action<HttpClientJsonOptions>(fun o ->
+                    Json.configureDefaultSerializerOptions Seq.empty o.SerializerOptions
+                )
+            )
+            // Use for pretty printing in logs
```
valbers, line 32: Nice 👌

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -989,8 +988,8 @@ and ScalarDef =
         abstract Name : string
         /// Optional scalar type description.
         abstract Description : string option
-        /// A function used to retrieve a .NET object from provided GraphQL query.
-        abstract CoerceInput : InputParameterValue -> obj option
+        /// A function used to retrieve a .NET object from provided GraphQL query or JsonElement variable.
+        abstract CoerceInput : InputParameterValue -> Result<obj, IGQLError list>
```
valbers, line 992: So this is the core change as I understand it (according to the PR's title).

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -627,7 +626,7 @@ and PlanningContext =
 
 /// A function type, which upon execution returns true if related field should
 /// be included in result set for the query.
-and Includer = ImmutableDictionary<string, obj> -> bool
+and Includer = ImmutableDictionary<string, obj> -> Result<bool, IGQLError list>
```
valbers, line 629: I think it would also make sense to use an `IReadOnlyDictionary<string, obj>` here like you did on line 1571.

src/FSharp.Data.GraphQL.Shared/TypeSystem.fs
```diff
@@ -2406,145 +2447,164 @@ module SchemaDefinitions =
...
             match Int32.TryParse(s, NumberStyles.Float, CultureInfo.InvariantCulture) with
-            | true, i -> Some i
-            | false, _ -> None
-        | InlineConstant (BooleanValue b) -> Some(if b then 1 else 0)
-        | _ -> None
+            | true, i -> Ok i
+            | false, _ -> getParseRangeError(destinationType, Int32.MinValue, Int32.MaxValue) s
+        | InlineConstant (BooleanValue b) -> Ok (if b then 1 else 0)
+        | InlineConstant value -> value.GetCoerceError destinationType
+        | _ -> Debug.Fail "Compiler bug, must not ever happen"; Unchecked.defaultof<_>
```
valbers, line 2467: not ever = never 🤓

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -94,12 +97,22 @@ type GQLProblemDetails =
             | :? IGQLErrorExtensions as ext -> ext.Extensions |> Skippable.ofValueOption
             | _ -> Skip
 
-        GQLProblemDetails.Create (error.Message, extensions)
+        let message =
+            match error with
+            | :? ICoerceGQLError as error -> error.VariableMessage + error.Message
```
valbers, line 102: Maybe it's a good idea to always have at least a whitespace character between both messages? So:
`error.VariableMessage + " " + error.Message`

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -94,12 +97,22 @@ type GQLProblemDetails =
...
     static member OfFieldError (path : FieldPath) (error : IGQLError) =
         let extensions =
             match error with
             | :? IGQLErrorExtensions as ext -> ext.Extensions |> Skippable.ofValueOption
             | _ -> Skip
 
-        GQLProblemDetails.Create (error.Message, path, extensions)
+        let message =
+            match error with
+            | :? ICoerceGQLError as error -> error.VariableMessage + error.Message
```
valbers, line 115: Here the same.

src/FSharp.Data.GraphQL.Server/Errors.fs
```diff
@@ -0,0 +1,114 @@
...
+        | Ok _ -> None
+        | Error err -> Some err)
+    |> Seq.toList
+
+let getObjectValues (object: IReadOnlyDictionary<string, Result<'t, IGQLError>>) =
+    object
+    |> Seq.map (fun kvp ->
+        match kvp.Value with
+        | Ok value -> KeyValuePair(kvp.Key, value)
+        | Error _ -> raise <| ArgumentException())
```
valbers, line 25: I'm not sure why an exception should be raised here...
Isn't this function just about selecting the `Result<,>`s which are of case `Ok`?
If the use case with throwing exception is the only valid use case, then I'd suggest to rename the function to something like `requireValuesWithNoErrors`.

src/FSharp.Data.GraphQL.Server/Errors.fs
```diff
@@ -0,0 +1,114 @@
...
+        let values = object |> getObjectValues
+        Ok values
+
+let getObjectErrorsList (object: IReadOnlyDictionary<string, Result<'t, IGQLError list>>) =
+    object
+    |> Seq.choose (fun kvp ->
+        match kvp.Value with
+        | Ok _ -> None
+        | Error err -> Some err)
+    |> Seq.toList
```
valbers, line 45: Are you sure you don't want to flatten the list before returning it? I mean:
```fsharp
|> Seq.collect id
|> Seq.toList
```

src/FSharp.Data.GraphQL.Server/Errors.fs
```diff
@@ -0,0 +1,114 @@
...
+
+let getObjectValuesList (object: IReadOnlyDictionary<string, Result<'t, IGQLError list>>) =
```
valbers, line 47: What do you think about having functions that require something contain the keyword "require" in their names? For example:
`let requireObjectValuesList ...`

src/FSharp.Data.GraphQL.Server/Errors.fs
```diff
@@ -0,0 +1,114 @@
...
+        let values = items |> getSeqValues
+        Ok values
+
+let getSeqErrorsList (items: Result<'t, IGQLError list> seq) =
+    items
+    |> Seq.choose (fun result ->
+        match result with
+        | Ok _ -> None
+        | Error err -> Some err)
+    |> Seq.toList
```
valbers, line 97: Here again, just as a reminder: you'd probably want to flatten the list here already.

src/FSharp.Data.GraphQL.Server/Errors.fs
```diff
@@ -0,0 +1,114 @@
...
+
+let getSeqValuesList (items: Result<'t, IGQLError list> seq) =
+    items
+    |> Seq.map (fun result ->
+        match result with
+        | Ok value -> value
+        | Error _ -> raise <| ArgumentException())
+    |> Seq.toArray
```
valbers, line 105: Suddently an array (instead of a list) here? 😅

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -130,21 +130,39 @@ module internal Gen =
 
 module internal ReflectionHelper =
 
+    let isParameterOptional (p: ParameterInfo) =
+        p.IsOptional
+        || p.ParameterType.FullName.StartsWith "Microsoft.FSharp.Core.FSharpOption`1"
+        || p.ParameterType.FullName.StartsWith "Microsoft.FSharp.Core.FSharpValueOption`1"
+
+    let isPrameterMandatory = not << isParameterOptional
```
valbers, line 138:
```suggestion
    let isParameterMandatory = not << isParameterOptional
```

tests/FSharp.Data.GraphQL.Tests/AstValidationTests.fs
```diff
@@ -2,10 +2,10 @@
 // Copyright (c) 2016 Bazinga Technologies Inc
 module FSharp.Data.GraphQL.Tests.AstValidationTests
 
+open Xunit
```
valbers, line 5: 🤷

tests/FSharp.Data.GraphQL.Tests/Relay/CursorTests.fs
```diff
@@ -0,0 +1,96 @@
...
+let ``Relay cursor works for types with nested fileds`` () =
+    let viewer =
+        { Id = "1"
+          Name = "Anonymous"
+          Widgets =
+            [ { Id = "1"; Name = "What's it" }
+              { Id = "2"; Name = "Who's it" }
+              { Id = "3"; Name = "How's it" } ] }
+
+    let getUser id = if viewer.Id = id then Some viewer else None
```
valbers:
```suggestion
    let tryGetUser id = if viewer.Id = id then Some viewer else None
```

valbers:
```suggestion
    let tryGetUser id = if viewer.Id = id then Some viewer else None
```

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/InputComplexTests.fs
```diff
@@ -0,0 +1,246 @@
...
+    )
+
+type TestInput =
+    { mand : string
+      opt1 : string option
+      opt2 : string option
+      optSeq : string option seq option
+      voptSeq : string option seq voption // string voption seq voption is too hard to implement
+      optArr : string option array option
+      voptArr : string option array voption } // string voption array voption is too hard to implement
```
valbers, line 44: 😂

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/InputEnumTests.fs
```diff
@@ -0,0 +1,142 @@
...
+    JsonSerializer.Serialize (arg, Json.serializerOptions)
+
+let stringifyInput = stringifyArg "input"
+
+type EnumTestType =
+    | Foo
+    | Bar
+
+let EnumTestType =
+    Define.Enum ("EnumTestType", [ Define.EnumValue ("Foo", EnumTestType.Foo); Define.EnumValue ("Bar", EnumTestType.Bar) ], "Test enum")
```
valbers, line 29: IMHO it's better readable when it's a multi-line definition instead of an one-liner.

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/InputListTests.fs
valbers, line 1: Excelent! 👍

valbers (approved):
Overall it looks good to me. It would be nice if you reacted to (some of) my comments, though.

src/FSharp.Data.GraphQL.Shared/Prolog.fs
xperiandri: It is not possible as I do not have the generic type argument. So the type must be instantiated dynamically

.editorconfig
xperiandri: I prefer `stroustrup` as `cramped` is hard to manage

samples/star-wars-api/JSON.fs
xperiandri: `configureDefaultSerializerOptions` became `unit` instead of options.
Is this what you mean?

valbers (approved):
LGTM

samples/star-wars-api/JSON.fs
valbers: Oh, ok, it looks like I missed it.

PR #439, opened 2023-08-27
Implemented `ScalarDefinition<'Primitive, 'Val>`, fixed built-in scalars input coercion

xperiandri:
* Added `ScalarDefinition<'Primitive, 'Val>`
* Changed `IDType` implementation to support only string and int64 Ids
* Fixed built-in scalars input coercion and added tests for that

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
+++ b/src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
@@ -21,14 +21,27 @@ module SchemaDefinitions =
         type InputValue with
 
             member inputValue.GetCoerceError(destinationType) =
-                let getMessage inputType value = $"Inline value '{value}' of type %s{inputType} cannot be converted to %s{destinationType}"
+                let getMessage inputType value = $"Inline value '{value}' of type %s{inputType} cannot be converted into %s{destinationType}"
                 let message =
                     match inputValue with
                     | IntValue value -> getMessage "integer" value
                     | FloatValue value -> getMessage "float" value
                     | BooleanValue value -> getMessage "boolean" value
                     | StringValue value -> getMessage "string" value
-                    | NullValue ->  $"Inline null value cannot be converted to {destinationType}"
+                    | NullValue ->  $"Inline value 'null' cannot be converted into {destinationType}"
+                    | EnumValue value ->  getMessage "enum" value
+                    | value -> raise <| NotSupportedException $"{value} cannot be passed as scalar input"
+                Error [{ new IGQLError with member _.Message = message }]
+
+            member inputValue.GetCoerceRangeError(destinationType, minValue, maxValue) =
+                let getMessage inputType value = $"Inline value '{value}' of type %s{inputType} cannot be converted into %s{destinationType} of range from {minValue} to {maxValue}"
+                let message =
+                    match inputValue with
+                    | IntValue value -> getMessage "integer" value
+                    | FloatValue value -> getMessage "float" value
+                    | BooleanValue value -> getMessage "boolean" value
+                    | StringValue value -> getMessage "string" value
+                    | NullValue ->  $"Inline value 'null' cannot be converted into {destinationType}"
                     | EnumValue value ->  getMessage "enum" value
                     | value -> raise <| NotSupportedException $"{value} cannot be passed as scalar input"
                 Error [{ new IGQLError with member _.Message = message }]
@@ -36,16 +49,18 @@ module SchemaDefinitions =
         type JsonElement with
 
             member e.GetDeserializeError(destinationType, minValue, maxValue ) =
-                Error [{ new IGQLError with member _.Message = $"Cannot deserialize JSON value '{e.GetRawText()}' into %s{destinationType} of range from {minValue} to {maxValue}" }]
+                let jsonValue = match e.ValueKind with JsonValueKind.String -> e.GetString() | _ -> e.GetRawText()
+                Error [{ new IGQLError with member _.Message = $"JSON value '{jsonValue}' of kind '{e.ValueKind}' cannot be deserialized into %s{destinationType} of range from {minValue} to {maxValue}" }]
 
             member e.GetDeserializeError(destinationType) =
-                Error [{ new IGQLError with member _.Message = $"Cannot deserialize JSON value '{e.GetRawText()}' into %s{destinationType}" }]
+                let jsonValue = match e.ValueKind with JsonValueKind.String -> e.GetString() | _ -> e.GetRawText()
+                Error [{ new IGQLError with member _.Message = $"JSON value '{jsonValue}' of kind '{e.ValueKind}' cannot be deserialized into %s{destinationType}" }]
 
         let getParseRangeError (destinationType, minValue, maxValue) value =
-            Error [{ new IGQLError with member _.Message = $"Cannot parse '%s{value}' into %s{destinationType} of range from {minValue} to {maxValue}" }]
+            Error [{ new IGQLError with member _.Message = $"Inline value '%s{value}' cannot be parsed into %s{destinationType} of range from {minValue} to {maxValue}" }]
 
         let getParseError destinationType value =
-            Error [{ new IGQLError with member _.Message = $"Cannot parse '%s{value}' into %s{destinationType}" }]
+            Error [{ new IGQLError with member _.Message = $"Inline value '%s{value}' cannot be paprsed into %s{destinationType}" }]
 
 
     open System.Globalization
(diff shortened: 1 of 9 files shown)


mickhansen:
Would it be possible to add some tests over `ValueObjectScalarDefinition` / `ValueObject`? 
I'm not quite seeing the purpose of the new definition, tests might help clarify that for me.

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/CoercionTests.fs
```diff
@@ -17,45 +18,97 @@ let private testCoercion graphQLType (expected: 't) actual =
     let result = (scalar.CoerceInput actual) |> Result.map (fun x -> downcast x)
     match result with
     | Ok x -> equals expected x
-    | Error _ -> raise (Exception(sprintf "Expected %A to be able to be coerced to %A" actual expected))
+    | Error _ -> Assert.Fail $"Expected %A{actual} to be able to be coerced to %A{expected}"
```
valbers, line 22: nice 👌

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/CoercionTests.fs
```diff
@@ -17,45 +18,97 @@ let private testCoercion graphQLType (expected: 't) actual =
...
+
+let private testCoercionError graphQLType (expectedErrorMeesage: string) actual =
```
valbers:
```suggestion
let private testCoercionError graphQLType (expectedErrorMessage: string) actual =
```

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/CoercionTests.fs
```diff
@@ -17,45 +18,97 @@ let private testCoercion graphQLType (expected: 't) actual =
...
 
 
 [<Fact>]
 let ``Int coerces input`` () =
+    testCoercion IntType 123 (Variable (JsonDocument.Parse "123").RootElement)
     testCoercion IntType 123 (InlineConstant (IntValue 123L))
-    testCoercion IntType 123 (InlineConstant (FloatValue 123.4))
-    testCoercion IntType 123 (InlineConstant (StringValue "123"))
+    testCoercionError IntType "JSON value '123.4' of kind 'Number' cannot be deserialized into integer of range from -2147483648 to 2147483647" (Variable (JsonDocument.Parse "123.4").RootElement)
+    testCoercionError IntType "Inline value '123,4' of type float cannot be converted into integer of range from -2147483648 to 2147483647" (InlineConstant (FloatValue 123.4))
```
valbers: What about this comma (i.e. `,`) here in the number? I think it would be better to have a dot representation which is "canonical" (we define floats with a dot). Besides, this test will probably behave differently on another machine with a culture that uses dots in numbers with decimal points.

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/CoercionTests.fs
```diff
@@ -17,45 +18,97 @@ let private testCoercion graphQLType (expected: 't) actual =
...
     testCoercion IntType 0 (InlineConstant (BooleanValue false))
+    testCoercionError IntType "JSON value 'enum' of kind 'String' cannot be deserialized into integer of range from -2147483648 to 2147483647" (Variable (JsonDocument.Parse "\"enum\"").RootElement)
+    testCoercionError IntType "Inline value 'enum' of type enum cannot be converted into integer of range from -2147483648 to 2147483647" (InlineConstant (EnumValue "enum"))
+
+[<Fact>]
+let ``Long coerces input`` () =
+    testCoercion LongType 123L (Variable (JsonDocument.Parse "123").RootElement)
+    testCoercion LongType 123L (InlineConstant (IntValue 123L))
+    testCoercionError LongType "JSON value '123.4' of kind 'Number' cannot be deserialized into integer of range from -9223372036854775808 to 9223372036854775807" (Variable (JsonDocument.Parse "123.4").RootElement)
+    testCoercionError LongType "Inline value '123,4' of type float cannot be converted into integer of range from -9223372036854775808 to 9223372036854775807" (InlineConstant (FloatValue 123.4))
```
valbers: here the comma thing again.

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/CoercionTests.fs
```diff
@@ -17,45 +18,97 @@ let private testCoercion graphQLType (expected: 't) actual =
...
     testCoercion StringType "false" (InlineConstant (BooleanValue false))
+    testCoercion StringType "enum" (Variable (JsonDocument.Parse "\"enum\"").RootElement)
+    testCoercion StringType "enum" (InlineConstant (EnumValue "enum"))
+
+[<Fact>]
+let ``ID coerces input`` () =
+    testCoercion IDType "123" (Variable (JsonDocument.Parse "123").RootElement)
+    testCoercion IDType "123" (InlineConstant (IntValue 123L))
+    testCoercionError IDType "JSON value '123.4' of kind 'Number' cannot be deserialized into identifier of range from -9223372036854775808 to 9223372036854775807" (Variable (JsonDocument.Parse "123.4").RootElement)
+    testCoercionError IDType "Inline value '123,4' of type float cannot be converted into identifier of range from -9223372036854775808 to 9223372036854775807" (InlineConstant (FloatValue 123.4))
```
valbers: comma (just so you don't forget).

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
```diff
@@ -21,31 +21,46 @@ module SchemaDefinitions =
...
+                let jsonValue = match e.ValueKind with JsonValueKind.String -> e.GetString() | _ -> e.GetRawText()
+                Error [{ new IGQLError with member _.Message = $"JSON value '{jsonValue}' of kind '{e.ValueKind}' cannot be deserialized into %s{destinationType}" }]
 
         let getParseRangeError (destinationType, minValue, maxValue) value =
-            Error [{ new IGQLError with member _.Message = $"Cannot parse '%s{value}' into %s{destinationType} of range from {minValue} to {maxValue}" }]
+            Error [{ new IGQLError with member _.Message = $"Inline value '%s{value}' cannot be parsed into %s{destinationType} of range from {minValue} to {maxValue}" }]
 
         let getParseError destinationType value =
-            Error [{ new IGQLError with member _.Message = $"Cannot parse '%s{value}' into %s{destinationType}" }]
+            Error [{ new IGQLError with member _.Message = $"Inline value '%s{value}' cannot be paprsed into %s{destinationType}" }]
```
valbers:
```suggestion
            Error [{ new IGQLError with member _.Message = $"Inline value '%s{value}' cannot be parsed into %s{destinationType}" }]
```

xperiandri:
Good point, I'll add.
The main idea is that you have a basic primitive like string and a wrapped value object like `type MyId = Id of Guid`.
So from the variable string comes to coercion function and then you perform custom processing.
If you define a scalar of `MyId` you will have to have a variable to be `MyId` on variables coercion, and on inline value parsing of AST. The first one is in theory more or less possible but the second one I do not know how to perform.

valbers:
I agree with @mickhansen regarding the question about the purpose of `ValueObjectScalarDefinition`, besides I'm struggling to find it a good naming. In the end, the definition of `ValueObjectScalarDefinition` doesn't imply any value object of sorts, it's simply a further generalization of `ScalarDefinition<'Val>` such that `type ScalarDefinition<'Val> = ValueObjectScalarDefinition<'Val, 'Val>`.
So I'd recommend to at least change the name to something like `type AsymmetricScalarDefinition<'CoercedValue, 'CoercedInput>`

xperiandri:
Probably it can become `ScalarDefinition'2` as it will not conflict with `ScalarDefinition'1` 🙂

valbers (changes requested):
Please address the open points and then I can approve your PR.

mickhansen:
> Good point, I'll add. The main idea is that you have a basic primitive like string and a wrapped value object like `type MyId = Id of Guid`. So from the variable string comes to coercion function and then you perform custom processing. If you define a scalar of `MyId` you will have to have a variable to be `MyId` on variables coercion, and on inline value parsing of AST. The first one is in theory more or less possible but the second one I do not know how to perform.

That sounds like a good usecase! Just wasn't familiar with the `ValueObject` terminology.

xperiandri:
@valbers your comments are resolved. Thanks for the very valuable feedback!

GBirkel (approved):

valbers (approved):
Looks good to me ™️

PR #440, opened 2023-09-11
Implemented error details as extensions

xperiandri:
Now errors look like
``` JSON
{
	"message": "mandatory",
	"path": [ "field", 0 ], // if field
	"extensions": {
		"kind": "Validation|InputCoercion|InputObjectValidation|Execution" // one of four kinds
		"path": [ "inputObjectField", 0 ], // if input object field, path within input object or array
		"variableName": "if variable",
		"variableType": "if variable",
		"argumentName": "if argument",
		"argumentType": "if argument",
		"objectType": "if object",
		"fieldType": "if field"
	}
}

### CODE REVIEW

File: src/FSharp.Data.GraphQL.Server/Execution.fs [hunk 1]
```diff
@@ -269,8 +269,9 @@ type StreamOutput =
 let private raiseErrors errs = AsyncVal.wrap <| Error errs
 
 /// Given an error e, call ParseError in the given context's Schema to convert it into
-/// a list of one or more <see herf="IGQLErrors">IGQLErrors</see>, then convert those to a list of <see href="GQLProblemDetails">GQLProblemDetails</see>.
-let private resolverError path ctx e = ctx.Schema.ParseError path e |> List.map (GQLProblemDetails.OfFieldError (path |> List.rev))
+/// a list of one or more <see herf="IGQLErrors">IGQLErrors</see>, then convert those
```
GBirkel:
```suggestion
/// a list of one or more <see href="IGQLErrors">IGQLErrors</see>, then convert those
```

GBirkel (approved):

src/FSharp.Data.GraphQL.Server/Exceptions.fs
```diff
@@ -0,0 +1,15 @@
+// The MIT License (MIT)
+// Copyright (c) 2016 Bazinga Technologies Inc
```
valbers, line 2: hmm... MIT or Copyright? 😅

src/FSharp.Data.GraphQL.Server/Execution.fs
```diff
@@ -642,9 +643,9 @@ let private executeSubscription (resultSet: (string * ExecutionInfo) []) (ctx: E
 let private compileInputObject (indef: InputObjectDef) =
     indef.Fields
     |> Array.iter(fun inputField ->
-        // TODO: Implement compilaiton cache to reuse for the same type
-        let errMsg = $"Input object '%s{indef.Name}': in field '%s{inputField.Name}': "
-        inputField.ExecuteInput <- compileByType errMsg inputField.TypeDef
+        // TODO: Implement compilation cache to reuse for the same type
+        let inputFieldTypeDef = inputField.TypeDef
+        inputField.ExecuteInput <- compileByType [ box inputField.Name ] Unknown (inputFieldTypeDef, inputField.TypeDef)
```
valbers: Why `(inputFieldTypeDef, inputField.TypeDef)` instead of `(inputFieldTypeDef, inputFieldTypeDef)`?

src/FSharp.Data.GraphQL.Server/ErrorTypes.fs
```diff
@@ -0,0 +1,124 @@
...
+    mutable InputSource : InputSource
+    InnerError : IGQLError
+    ErrorKind : ErrorKind
+    Path : FieldPath
+    FieldErrorDetails : ObjectFieldErrorDetails voption
+} with
+
+    interface IGQLError with
+
+        member this.Message = this.InnerError.Message
```
valbers, line 80: The wrapper has its own `InputSource`, `ErrorKind`, `Path` and `FieldErrorDetails`. Why doesn't it have its own message?

src/FSharp.Data.GraphQL.Shared/ValidationTypes.fs
```diff
@@ -45,5 +53,8 @@ module GQLValidator =
 type AstError =
 
     static member AsResult(message : string, ?path : FieldPath) =
-        [ { Message = message; Path = path |> Skippable.ofOption |> Skippable.map List.rev; Locations = Skip; Extensions = Skip  } ]
+        [
+            let extensions = Dictionary<string, obj>() |> GQLProblemDetails.SetErrorKind ErrorKind.Validation
+            { Message = message; Path = path |> Skippable.ofOption |> Skippable.map List.rev; Locations = Skip; Extensions = Include extensions }
```
valbers: IMHO it's better readable when it's multi-line instead of everything in one line 🙂

src/FSharp.Data.GraphQL.Shared/Validation.fs
```diff
@@ -934,7 +934,7 @@ module Ast =
             | VariableName varName ->
                 if variableDefinitions |> Set.contains varName
                 then Success
-                else AstError.AsResult($"A variable '%s{varName}' is referenced in an argument '%s{arg.Name}' of directive '%s{directive.Name}' of field with alias or name '%O{path.Head}', but that variable is not defined in the operation.", path)
+                else AstError.AsResult($"Variable '%s{varName}' is referenced in an argument '%s{arg.Name}' of directive '%s{directive.Name}' of field with alias or name '%O{path.Head}', but that variable is not defined in the operation.", path)
```
valbers: The previous form was actually correct and kind of more conversational 🙂 . This is a subtle difference in English. By putting the indefinite article "a" in front of "variable" it means "something which is not known", which is confirmed in the "but that variable is not defined in the operation" part of the phrase. If you remove the "a" it's still correct, but it loses that subtle hint given right at the beginning about it being an unknown variable (because it was not defined).

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -9,15 +9,16 @@ open System.Collections.Generic
...
 open System.Text.Json
 open FsToolkit.ErrorHandling
 
+open FSharp.Data.GraphQL
 open FSharp.Data.GraphQL.Ast
 open FSharp.Data.GraphQL.Errors
 open FSharp.Data.GraphQL.Types
 open FSharp.Data.GraphQL.Types.Patterns
 open FSharp.Data.GraphQL.Validation
+open FSharp.Data.GraphQL
```
valbers, line 19:
```suggestion
```

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -189,140 +204,196 @@ let rec internal compileByType (errMsg : string) (inputDef : InputDef) : Execute
...
+        isNullable
+        inputObjectPath
+        (objectFieldErrorDetails: ObjectFieldErrorDetails voption)
+        (originalTypeDef, typeDef) (varDef : VarDef)
+        (input : JsonElement)
+    : Result<obj, IGQLError list> =
+
+    let createVariableCoercionError message =
+        Error [ {
+            InputSource = Variable varDef
```
valbers: I find it more helpful for the reader if you be a little bit more explicit about the type you're creating here, like:
```suggestion
            CoercionError.InputSource = Variable varDef
```
because there's another type `CoercionErrorWrapper` which looks almost just like this one.

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -189,140 +204,196 @@ let rec internal compileByType (errMsg : string) (inputDef : InputDef) : Execute
...
+
+    let createNullError typeDef =
+        let message =
+            match objectFieldErrorDetails with
+            | ValueSome details -> $"Non-nullable field '%s{details.FieldDef.Value.Name}' expected value of type '%s{string typeDef}', but got 'null'."
+            | ValueNone -> $"Non-nullable variable '$%s{varDef.Name}' expected value of type '%s{string typeDef}', but got 'null'."
+        createVariableCoercionError message
+
+    let mapInputError varDef inputObjectPath (objectFieldErrorDetails: ObjectFieldErrorDetails voption) (err : IGQLError) : IGQLError = {
+        InnerError = err
```
valbers, line 299: On that note, here one would also have:
```suggestion
        CoercionErrorWrapper.InnerError = err
```

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -189,140 +204,196 @@ let rec internal compileByType (errMsg : string) (inputDef : InputDef) : Execute
...
     | _ ->
         Debug.Fail "Unexpected InputDef"
         failwithf "Unexpected value of inputDef: %O" inputDef
 
-let rec internal coerceVariableValue isNullable typedef (vardef : VarDef) (input : JsonElement) (errMsg : string) : Result<obj, IGQLError list> =
-    match typedef with
+let rec internal coerceVariableValue
+        isNullable
+        inputObjectPath
+        (objectFieldErrorDetails: ObjectFieldErrorDetails voption)
```
valbers, line 277: It's probably not going to be a big deal anyway, but I'm thinking if it'd make sense memory/performance-wise to have a function here that would return the error details in case of error instead of always passing the concrete error details even when there are no errors, which will be the most taken runtime path anyway (once the consumer app is working, then you wouldn't get these types of errors in the normal case).

valbers (changes requested):
It looks good, but please check my comment about passing an error details object as an argument vs passing a function that will return that on demand.

src/FSharp.Data.GraphQL.Server/ErrorTypes.fs
xperiandri: Because a custom error that come from the library user can implement `IGQLErrorExtensions` and in that case it will have more properties. And in order to grab them I need to store them somewhere.
So I decided to go with decorator pattern and delay this to the very end

src/FSharp.Data.GraphQL.Server/Execution.fs
xperiandri: It is a typo

src/FSharp.Data.GraphQL.Shared/ValidationTypes.fs
xperiandri: What about this way? Or just a property on each line is better?
```suggestion
            let path = path |> Skippable.ofOption |> Skippable.map List.rev
            { Message = message; Path = path; Locations = Skip; Extensions = Include extensions }
```

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri: How do you think we can capture that?

src/FSharp.Data.GraphQL.Shared/ValidationTypes.fs
valbers: I don't know if it's just me, but I find it better readable one line per property:
```suggestion
            { Message = message
              Path = path |> Skippable.ofOption |> Skippable.map List.rev
              Locations = Skip
              Extensions = Include extensions }
```

src/FSharp.Data.GraphQL.Server/Values.fs
valbers: `getObjectFieldErrorDetails: unit -> ObjectFieldErrorDetails voption`

and when calling `coerceVariableValue`:
`coerceVariableValue ... (fun () -> ...) ...`

xperiandri:
@valbers fixed major comments including variable error messages

valbers (approved):

PR #446, opened 2023-10-17
Added `FieldDef<'Val, 'Res>` to support resolver changing middlewares

xperiandri:
In order to call `Define.Field<'Val>(....).WithAuthorizationPolicy("policy")` I need to know not only `'Res` which is an object on which `FieldDef<'Res>` is defined but also a `'Val` which is a field type.
However when I introduced `FieldDef<'Val, 'Res>` F# cannot determine the override between `Define.Object` that gets a list of field and a labmda that return a list of fields. So I had to rename `Define.Object` overrie to `Define.ObjectRec` and `Define.Interface` override to `Define.ObjectRec`.
I have not renamed `Define.InputObject` as it accepts a parameter of list of another interface

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
+++ b/src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
@@ -666,7 +666,8 @@ module SchemaDefinitions =
               DeprecationReason = deprecationReason }
 
         /// <summary>
-        /// Creates GraphQL custom output object type. It can be used as a valid output but not an input object
+        /// Creates a GraphQL custom output object type that has a field of a type that references this object type recursively.
+        /// It can be used as a valid output but not as an input object
         /// (see <see cref="InputObject"/> for more details).
         /// </summary>
         /// <param name="name">Type name. Must be unique in scope of the current schema.</param>
@@ -680,7 +681,7 @@ module SchemaDefinitions =
         /// <param name="isTypeOf">
         /// Optional function used to determine if provided .NET object instance matches current object definition.
         /// </param>
-        static member Object(name : string, fieldsFn : unit -> FieldDef<'Val> list, ?description : string,
+        static member ObjectRec(name : string, fieldsFn : unit -> FieldDef<'Val> list, ?description : string,
                              ?interfaces : InterfaceDef list, ?isTypeOf : obj -> bool) : ObjectDef<'Val> =
             upcast { ObjectDefinition.Name = name
                      Description = description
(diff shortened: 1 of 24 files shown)


GBirkel:
I'm afraid this one may be beyond my skill to review.  I'm not sure what the long-term consequences of changing `Object` to `ObjectRec` would be.

xperiandri:
Anyway we do breaking changes in v2.0
So the main difference is that you need to call `Define.ObjectRec` to define recursive GraphQL type definitions.
Also there is a question if need to rename `Define.InputObject` override to `Define.InputObjectRec` to be consistent or leave it as is to be simple?

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
```diff
@@ -665,7 +665,8 @@ module SchemaDefinitions =
               DeprecationReason = deprecationReason }
 
         /// <summary>
-        /// Creates GraphQL custom output object type. It can be used as a valid output but not an input object
+        /// Creates GraphQL custom output object type that has a field of type that refernces this object type recurively.
```
valbers:
```suggestion
        /// Creates a GraphQL custom output object type that has a field of a type that references this object type recursively.
```

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
```diff
@@ -665,7 +665,8 @@ module SchemaDefinitions =
...
+        /// It can be used as a valid output but not an input object
```
valbers:
```suggestion
        /// It can be used as a valid output but not as an input object
```

valbers (changes requested):
I get it.
In any case, I find it a good idea to generally avoid method overloads as F# doesn't support function overloads (although class members may have overloads, just like in our current situation -- which is causing the problem that you're trying to fix), and therefore not having overloads feels more like "the F# way", so I welcome this change already on this level, even if the offending overload were not causing any problems yet.

However, I'm not sure if I agree with the new name. `ObjectRec` is effectively not by definition recursive, but it _can_ be used for a recursive definition.
That, which `ObjectRec` actually is, is an `Object` which takes a function for defining the fields instead of a concrete list of fields.

So, following the nomenclature of the F# core lib, I'd prefer for it to be named something like the [`defaultWith`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html#defaultWith) function of the `Option` module.

In that case, that could be `ObjectHavingFieldsWith` (instead of `ObjectRec`).
What do you think?

GBirkel (approved):
Have the changes requested by valbers been addressed?

xperiandri:
@njlr give your opinion, please

valbers:
@xperiandri could you please do less force-pushes with commit amendments? It makes it hard to follow what you actually changed with each commit amendment. I mean: once in a while it's ok, but I see this now very often.

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
```diff
@@ -800,15 +801,14 @@ module SchemaDefinitions =
         /// <param name="description">Optional field description. Usefull for generating documentation.</param>
         /// <param name="args">Optional list of arguments used to parametrize field resolution.</param>
         /// <param name="deprecationReason">If set, marks current field as deprecated.</param>
-        static member AutoField(name : string, typedef : #OutputDef<'Res>, ?description: string, ?args: InputFieldDef list, ?deprecationReason: string) : FieldDef<'Val> =
+        static member AutoField(name : string, typedef : OutputDef<'Res>, ?description: string, ?args: InputFieldDef list, ?deprecationReason: string) : FieldDef<'Val, 'Res> =
```
Thorium, line 804: Was the # removed on purpose?

xperiandri: I either got a warning or error because of this

valbers (approved):
I gave you my two cents with regard to the naming of `ObjectRec`, but it seems like you don't agree with it. I don't want to die on that hill, so I'm approving this PR.

xperiandri:
I just want to check again if it still the issue.
And I cannot decide on naming 🙂

xperiandri:
@valbers I implemented an alternative solution #492

PR #448, opened 2023-10-19
Changed `raise` to calling `Reraise()` extension method to preserve the call stack

xperiandri:
@valbers do we need tests for this?


valbers:
> @valbers do we need tests for this?

Tests are of course always good, but I wouldn't require it here at this point. If you tested it locally it should suffice for now, I guess.

valbers (approved):

GBirkel (approved):

PR #450, opened 2023-10-29
Fixed packaging and publishing NuGet packages

xperiandri:
1. Fixed `The 'NoBuild' property was set to true but the 'Build' target was invoked. ` error for the `Relay` project
2. Fixed ability to skip already published packages and continue with not yet published using `--skip-duplicate` parameter

### DIFF
--- a/build/DotNet.fs
+++ b/build/DotNet.fs
@@ -27,7 +27,9 @@ module DotNet =
         if Environment.isUnix then
             Environment.environVar "HOME" @@ ".dotnet"
         else
-            Environment.environVar "LocalAppData" @@ "Microsoft" @@ "dotnet"
+            Environment.environVar "LocalAppData"
+            @@ "Microsoft"
+            @@ "dotnet"
 
     /// <summary>
     /// .NET Core SDK default install directory (set to default SDK installer paths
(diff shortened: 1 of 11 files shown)


build/DotNet.fs
```diff
@@ -27,7 +27,9 @@ module DotNet =
         if Environment.isUnix then
             Environment.environVar "HOME" @@ ".dotnet"
         else
-            Environment.environVar "LocalAppData" @@ "Microsoft" @@ "dotnet"
+            Environment.environVar "LocalAppData"
+            @@ "Microsoft"
+            @@ "dotnet"
```
mickhansen, line 32: Did any logic in this file change, or is it only formatting changes?

mickhansen (approved):
I'm not too well versed in how packaging works, but overall LGTM.

GBirkel (approved):

valbers (approved):
Fine by me

build/DotNet.fs
xperiandri: Only formatting.

Only this added
https://github.com/fsprojects/FSharp.Data.GraphQL/blob/76af489fe2ef701754827d34e38a02bab56021e2/build/NuGet.fs#L576-L578

PR #456, opened 2024-02-10
Implemented normalizing optionals on input objects

xperiandri:
@valbers I've implemented minimal tests. Do a brief check, please

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -73,5 +73,6 @@
     <PackageReference Update="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.*" />
     <PackageReference Update="Microsoft.NETCore.Platforms" Version="6.0.*" />
     <PackageReference Update="Newtonsoft.Json" Version="13.*" />
+    <PackageReference Update="Validus" Version="4.1.*" />
   </ItemGroup>
 </Project>
--- a/src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
+++ b/src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
@@ -130,10 +130,13 @@ module internal Gen =
 
 module internal ReflectionHelper =
 
+    let [<Literal>] OptionTypeName = "Microsoft.FSharp.Core.FSharpOption`1"
+    let [<Literal>] ValueOptionTypeName = "Microsoft.FSharp.Core.FSharpValueOption`1"
+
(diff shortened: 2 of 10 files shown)


Packages.props
```diff
@@ -73,5 +73,6 @@
     <PackageReference Update="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.*" />
     <PackageReference Update="Microsoft.NETCore.Platforms" Version="6.0.*" />
     <PackageReference Update="Newtonsoft.Json" Version="13.*" />
+    <PackageReference Update="Validus" Version="4.1.*" />
```
valbers, line 76: what about `FsToolkit.ErrorHandling`?

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -130,10 +130,13 @@ module internal Gen =
 
 module internal ReflectionHelper =
 
+    let [<Literal>] OptionTypeName = "Microsoft.FSharp.Core.FSharpOption`1"
```
valbers, line 133: Good 👍

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -312,6 +322,16 @@ let rec internal coerceVariableValue
             | Ok null when isNullable -> Ok null
             // TODO: Capture position in the JSON document
             | Ok null -> createNullError originalTypeDef
+            | Ok value when not isNullable ->
+                let ``type`` = value.GetType()
```
valbers, line 326: it's probably a matter of taste, but I find the name
```
``type``
```
rather cumbersome. The backticks trick is a genius move for (some, mostly unit-test) methods, but as a variable name, as I said, it's IMHO cumbersome.
I guess `type_` should also do the trick.
Just my two cents; do as you wish 🙂

valbers (approved):

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri: In the other code bases I saw `ty` and `typ` but it looks bad as for me 🙂

Packages.props
xperiandri: It is referenced transitively implicitly. Do you think we need to reference it explicitly?

PR #462, opened 2024-02-21
Fixed inline nullable null argument error

### LINKED ISSUE #461: 2.0.0-beta: Setting a `Nullable` input parameter to `null` fails
### Description

Queries and probably also mutations (but I did not check mutations) no longer support `Nullable` for input parameters when setting the input parameter to `null`. Setting an input parameter to `null` will always result in failing to convert it to the inner type of the `Nullable`. 

I think it goes wrong somewhere in file `Values.fs` `compileByType` as I see that `originalInputDef` is not used in the branch for `Scalar`. I don't understand the code well enough to figure out how it is supposed to work though. 

### Repro steps

Please see the following test that can be added on line 220 in `ExecutionTests.fs`:
```f#
[<Fact>]
let ``Execution handles basic tasks: correctly handles null arguments`` () =
    let query = """query Example {
        b(numArg: null, stringArg: null)
      }"""
    let data = { Num = None; Str = None }
    let Type =
        Define.Object("Type",
            [ Define.Field("b", Nullable StringType, "", [ Define.Input("numArg", Nullable IntType); Define.Input("stringArg", Nullable StringType) ],
                 fun ctx value ->
                     value.Num <- ctx.TryArg("numArg")
                     value.Str <- ctx.TryArg("stringArg")
                     value.Str) ])

    let result = sync <| Executor(Schema(Type)).AsyncExecute(parse query, data)
    ensureDirect result <| fun data errors -> empty errors
    equals None data.Num
    equals None data.Str
```

### Expected behavior

Setting an input parameter to `null` makes it `None`.

### Actual behavior

Setting a `Nullable IntType` (or any other inner type) input parameter to `null` causes this error: `Inline value 'null' cannot be converted into integer`.

### Known workarounds

No known workarounds.

### Related information 

Version 2.0.0-beta of the nuget packages.

xperiandri:
Fixes #461 

Co-authored-by: @panthus

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Values.fs
+++ b/src/FSharp.Data.GraphQL.Server/Values.fs
@@ -256,7 +256,10 @@ let rec internal compileByType (inputObjectPath: FieldPath) (inputSource : Input
         match innerDef with
         | InputObject inputObjDef -> inputObjDef.ExecuteInput <- inner
         | _ -> ()
-        fun value variables -> inner value variables
+        fun value variables ->
+            match value with
+            | NullValue -> Ok null
+            | _ -> inner value variables
 
     | Enum enumDef ->
         fun value variables ->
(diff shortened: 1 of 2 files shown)


valbers (approved):
LGTM

PR #466, opened 2024-03-24
Migrated to .NET 8 and updated all NuGet packages

xperiandri:
Also fixes `System.InvalidOperationException: This JsonSerializerOptions instance is read-only or has already been used in serialization or deserialization` when `configure` options function is passed to `AddGraphQLOptions`
And adds more overloads to `AddGraphQLOptions`


Directory.Build.props
```diff
@@ -18,7 +18,7 @@
     <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(MSBuildThisFileDirectory)'))))</RepoRoot>
     <RepositoryUrl>https://github.com/fsprojects/FSharp.Data.GraphQL</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
-    <Version>2.0.0</Version>
+    <Version>2.1.0</Version>
```
valbers: That would be a major change because projects targeting .net6 and referencing FSharp.Data.GraphQL version `2.0.0` will be in for a surprise when they update it to `2.1.0` thinking not much about it.
We know this from Microsoft and it's just an annoying behavior.

valbers (changes requested):
I think we have basically two options here:
  a) Keep supporting .NET 6 and go from version `2.0.0` to `2.1.0`;
  b) Drop support for .NET 6 and go from version `2.0.0` to `3.0.0`.

Directory.Build.props
xperiandri: New push has 3.0.0 in that place

PR #471, opened 2024-04-05
Improving server exception logging by including exception in log message

valbers:
There were two problems before:

1. The error message itself was not being logged, but only a mention that there was an error;
2. It was not possible to access the possibly original exception which led to the request processing error. However, it's extremely useful to log the error with the whole exception information including the stack trace of where the exception was thrown.

### Before
before-srv-ex-logging

### After
after-srv-ex-logging

  However, these changes also introduce a breaking change as `GQLProblemDetails` now has an additional member holding the possible exception which generated the error.

   `GQLProblemDetails` is public and could be in use by code using FSharp.Data.GraphQL.

### DIFF
--- a/src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
+++ b/src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
@@ -25,9 +25,9 @@ module HttpHandlers =
 
     let rec private moduleType = getModuleType <@ moduleType @>
 
-    let ofTaskIResult ctx (taskRes: Task<IResult>) : HttpFuncResult = task {
+    let ofTaskIResult ctx (taskRes : Task<IResult>) : HttpFuncResult = task {
         let! res = taskRes
-        do! res.ExecuteAsync(ctx)
+        do! res.ExecuteAsync (ctx)
         return Some ctx
     }
 
@@ -41,28 +41,28 @@ module HttpHandlers =
 
         let logger = sp.CreateLogger moduleType
 
-        let options = sp.GetRequiredService<IOptionsMonitor<GraphQLOptions<'Root>>>()
+        let options = sp.GetRequiredService<IOptionsMonitor<GraphQLOptions<'Root>>> ()
 
         let toResponse { DocumentId = documentId; Content = content; Metadata = metadata } =
 
             let serializeIndented value =
                 let jsonSerializerOptions = options.Get(IndentedOptionsName).SerializerOptions
-                JsonSerializer.Serialize(value, jsonSerializerOptions)
+                JsonSerializer.Serialize (value, jsonSerializerOptions)
 
             match content with
-            | Direct(data, errs) ->
-                logger.LogDebug(
+            | Direct (data, errs) ->
+                logger.LogDebug (
                     $"Produced direct GraphQL response with documentId = '{{documentId}}' and metadata:\n{{metadata}}",
                     documentId,
                     metadata
                 )
 
                 if logger.IsEnabled LogLevel.Trace then
-                    logger.LogTrace($"GraphQL response data:\n:{{data}}", serializeIndented data)
+                    logger.LogTrace ($"GraphQL response data:\n:{{data}}", serializeIndented data)
 
-                GQLResponse.Direct(documentId, data, errs)
-            | Deferred(data, errs, deferred) ->
-                logger.LogDebug(
+                GQLResponse.Direct (documentId, data, errs)
+            | Deferred (data, errs, deferred) ->
+                logger.LogDebug (
                     $"Produced deferred GraphQL response with documentId = '{{documentId}}' and metadata:\n{{metadata}}",
                     documentId,
                     metadata
@@ -71,41 +71,32 @@ module HttpHandlers =
                 if logger.IsEnabled LogLevel.Debug then
                     deferred
                     |> Observable.add (function
-                        | DeferredResult(data, path) ->
-                            logger.LogDebug(
-                                "Produced GraphQL deferred result for path: {path}",
-                                path |> Seq.map string |> Seq.toArray |> Path.Join
-                            )
+                        | DeferredResult (data, path) ->
+                            logger.LogDebug ("Produced GraphQL deferred result for path: {path}", path |> Seq.map string |> Seq.toArray |> Path.Join)
 
                             if logger.IsEnabled LogLevel.Trace then
-                                logger.LogTrace(
-                                    $"GraphQL deferred data:\n{{data}}",
-                                    serializeIndented data
-                                )
-                        | DeferredErrors(null, errors, path) ->
-                            logger.LogDebug(
-                                "Produced GraphQL deferred errors for path: {path}",
-                                path |> Seq.map string |> Seq.toArray |> Path.Join
-                            )
+                                logger.LogTrace ($"GraphQL deferred data:\n{{data}}", serializeIndented data)
+                        | DeferredErrors (null, errors, path) ->
+                            logger.LogDebug ("Produced GraphQL deferred errors for path: {path}", path |> Seq.map string |> Seq.toArray |> Path.Join)
 
                             if logger.IsEnabled LogLevel.Trace then
-                                logger.LogTrace($"GraphQL deferred errors:\n{{errors}}", errors)
-                        | DeferredErrors(data, errors, path) ->
-                            logger.LogDebug(
+                                logger.LogTrace ($"GraphQL deferred errors:\n{{errors}}", errors)
+                        | DeferredErrors (data, errors, path) ->
+                            logger.LogDebug (
                                 "Produced GraphQL deferred result with errors for path: {path}",
                                 path |> Seq.map string |> Seq.toArray |> Path.Join
                             )
 
                             if logger.IsEnabled LogLevel.Trace then
-                                logger.LogTrace(
+                                logger.LogTrace (
                                     $"GraphQL deferred errors:\n{{errors}}\nGraphQL deferred data:\n{{data}}",
                                     errors,
                                     serializeIndented data
                                 ))
 
-                GQLResponse.Direct(documentId, data, errs)
+                GQLResponse.Direct (documentId, data, errs)
             | Stream stream ->
-                logger.LogDebug(
+                logger.LogDebug (
                     $"Produced stream GraphQL response with documentId = '{{documentId}}' and metadata:\n{{metadata}}",
                     documentId,
                     metadata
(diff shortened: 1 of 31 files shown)


src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -11,6 +11,7 @@ type internal FieldPath = obj list
 
 type IGQLError =
     abstract member Message : string with get
+    abstract member Exception : (Exception option) with get
```
xperiandri: Maybe we just introduce one more interface?
```
type IGQLErrorException =
    abstract member Exception : Exception voption with get
```

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -127,50 +134,57 @@ type GQLProblemDetails = {
         | :? IReadOnlyDictionary<string, obj> as extensions -> extensions
         | _ -> ReadOnlyDictionary<string, obj> mutableExtensions
 
-    static member internal CreateWithKind (message, kind : ErrorKind, ?path) = {
+    static member internal CreateWithKind (message : string, ex : Exception option,  kind : ErrorKind, ?path) = {
```
xperiandri: Maybe we make them `voption` and `[<Optional>]`?
```suggestion
    static member internal CreateWithKind (message : string, kind : ErrorKind, [<Optional>] ex, [<Optional>] path) = {
```

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
xperiandri, line 1: If we introduce new interface this change will not be needed

tests/FSharp.Data.GraphQL.Tests/AstValidationTests.fs
```diff
@@ -11,8 +11,8 @@ open FSharp.Data.GraphQL.Types
 
 module internal GQLProblemDetails =
 
-    let CreateValidation message = GQLProblemDetails.CreateWithKind (message, Validation)
-    let CreateValidationFor path message = GQLProblemDetails.CreateWithKind (message, Validation, path)
+    let CreateValidation message = GQLProblemDetails.CreateWithKind (message, None, Validation)
+    let CreateValidationFor path message = GQLProblemDetails.CreateWithKind (message, None, Validation, path)
```
xperiandri: And with optionals we can place ex at the end and not specify here

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: But how would it be used? How would I access the exception later like I do when `IGQLError` has just another member? I think there can be only one interface in the end.

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: Fine by me.

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
valbers: Ok... But I still don't get how introducing a new interface will then actually solve the problem. Where will this interface be used? 🤔

tests/FSharp.Data.GraphQL.Tests/AstValidationTests.fs
valbers: I thought about that possibility too (i.e.: just append the exception parameter to the list of parameters). But the reason why I put the exception argument in the middle is because the message and the exception are tightly related, almost as if they were two members of the same type. So it makes more sense to have them one next to the other than separated by other parameters.

src/FSharp.Data.GraphQL.Shared/Errors.fs
xperiandri: Match if interface is implemented

src/FSharp.Data.GraphQL.Shared/SchemaDefinitions.fs
xperiandri: Then we can introduce more overloads

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: Although: I just rewrote my changes according to your suggestions regarding the extra interface and, with it, this change at `CreateWithKind` isn't necessary anymore.

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: Done.

tests/FSharp.Data.GraphQL.Tests/AstValidationTests.fs
valbers: Not necessary anymore with the now extra `IGQLExceptionError` interface.

valbers:
@xperiandri I rewrote my changes according to your suggestion regarding the extra interface. Please kindly review the current changes.

src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
```diff
@@ -139,11 +139,23 @@ module HttpHandlers =
...
-                    documentId,
-                    metadata
-                )
+                match errs |> List.choose (fun x -> x.Exception |> Option.ofValueOption) |> List.tryHead with
+                | Some ex ->
+                    logger.LogError(
+                        ex,
+                        "Error while processing request that generated response with documentId '{documentId}'",
+                        documentId
+                    )
```
xperiandri: Shouldn't we just call `logger.LogError` as many times as how many exceptions we have?

xperiandri (commented):
Just one suggestion

src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
valbers: Done.

valbers:
@xperiandri there's a commit for applying fantomas formatting to `HttpHandlers.fs` but I find it suspicious. Do you have an idea why the formatting is so different now?

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -216,7 +257,15 @@ type GQLProblemDetails = {
...
             | _ -> error.Message
 
-        GQLProblemDetails.Create (message, path, extensions)
+        match error with
+        | :? IGQLExceptionError as exceptionError ->
+            match extensions with
+            | Include x ->
+                GQLProblemDetails.CreateOfException(exceptionError.Message, exceptionError.Exception, path, extensions = x)
+            | Skip ->
+                GQLProblemDetails.CreateOfException(exceptionError.Message, exceptionError.Exception, path)
```
xperiandri: We can make it simpler. Anyway `option` is created under the hood
```suggestion
            let x = extensions |> Skippable.toOption
            GQLProblemDetails.CreateOfException(exceptionError.Message, exceptionError.Exception, path, ?extensions = x)
```

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -200,7 +233,15 @@ type GQLProblemDetails = {
...
```
xperiandri: The same as below

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -187,7 +212,15 @@ type GQLProblemDetails = {
...
             | _ -> error.Message
 
-        GQLProblemDetails.Create (message, extensions)
+        match error with
+        | :? IGQLExceptionError as exceptionError ->
+            match extensions with
+            | Include x ->
+                GQLProblemDetails.CreateOfException(exceptionError.Message, exceptionError.Exception, extensions = x)
+            | Skip ->
+                GQLProblemDetails.CreateOfException(exceptionError.Message, exceptionError.Exception)
```
xperiandri: The same as below

xperiandri (commented):
@valbers it is something bad on your side. I've just formatted it on my side and it is similar to what it was before your formatting.
Try restore the latest Fantomas with tool restore in this repo

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: This doesn't work. Believe me, I tried this 🙂 . It doesn't expect an `option`, although the parameter is `option` under the hood. It's apparently a characteristic of the `?extensions` syntax.

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: It's not possible. See my previous comment.

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: It's not possible. See my previous comment.

src/FSharp.Data.GraphQL.Shared/Errors.fs
xperiandri: I've made it work in my code but I implemented a better change

xperiandri: See PR to your branch

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: Done in your PR. Closing comment

src/FSharp.Data.GraphQL.Shared/Errors.fs
valbers: Done in your PR. Closing converstaion.

valbers:
@xperiandri waiting for your approving review.

src/FSharp.Data.GraphQL.Shared/Errors.fs
```diff
@@ -129,48 +140,71 @@ type GQLProblemDetails = {
...
+    static member Create (message, [<Optional>] extensions) = {
         Message = message
+        Exception = ValueNone
         Path = Skip
         Locations = Skip
         Extensions = extensions
     }
 
-    static member Create (message, path, ?extensions : IReadOnlyDictionary<string, obj>) = {
+    static member CreateOfException (message : string, ex : Exception, [<Optional>] path : FieldPath Skippable, [<Optional>] extensions : IReadOnlyDictionary<string, obj> Skippable) = {
```
xperiandri: Shouldn't we call it just `Create`? As in my mind if you emphasize that it is `CreateOfException` it is something like you read the Exception fields into the new object.
While adding just an input parameter have no additional cognitive load

xperiandri (commented):
What do you think?

src/FSharp.Data.GraphQL.Shared/Errors.fs
xperiandri: @njlr @nikhedonia @cmeeren @pkese any thoughts?

valbers: Ok. It's just that I tend to avoid overloads, because I like to have a glance at the code and know exactly which function/method is being called 😄 .
But yeah, I can rename it to `Create`. This way it blends in with the other overloads.

cmeeren: I don't use this library yet, but sure, I definitely prefer overloads in these cases, for several reasons I don't care articulating at the moment.

valbers: Done.

src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
```diff
@@ -115,48 +106,59 @@ module HttpHandlers =
...
+                else
+                    logger.LogWarning (
+                        ("Produced request error GraphQL response with:\n"
+                         + "- documentId: '{documentId}'\n"
+                         + "- error(s):\n    {requestError}\n"
+                         + "- metadata:\n    {metadata}\n"),
+                        documentId,
+                        errs,
+                        metadata
+                    )
```
xperiandri: Maybe we just iter over once without allocating new lists?
```suggestion
                let notHasException =
                    errs
                    |> Seq.map (fun err ->
                        err.Exception |> ValueOption.iter (fun ex ->
                            logger.LogError (ex, "Error while processing request that generated response with documentId '{documentId}'", documented)
                        err.Exception.IsNone
                    |> Seq.fold (fun acc notExn -> acc && notExn) true
                if notHasException then
                    logger.LogWarning (
                        ("Produced request error GraphQL response with:\n"
                         + "- documentId: '{documentId}'\n"
                         + "- error(s):\n    {requestError}\n"
                         + "- metadata:\n    {metadata}\n"),
                        documentId,
                        errs,
                        metadata
                    )
```

xperiandri: Probably fold can be replaced with `Seq.all id`

xperiandri (commented):
One last thing 🙂

src/FSharp.Data.GraphQL.Server.AspNetCore/Giraffe/HttpHandlers.fs
valbers: I call this "the Tetris effect in functional programming" 😄

valbers: Done.

xperiandri (approved):

xperiandri:
Cool! Lots of thanks for the contribution!

Issue #472, opened 2024-04-06
Nested InputObjects don't get populated


pkese:
I've added a small random generator query to star-wars-api example:

```fsharp
    let randoms =
        let inputs = [
            Define.Input ("count", IntType)
            Define.Input ("range", Nullable (Define.InputObject("RandomRange", [
                Define.Input ("min", IntType)
                Define.Input ("max", IntType)
            ])))
        ]
        let renderRandoms (ctx:ResolveFieldContext) (r:Root) =
            // printfn "Random called with count:%A and range:%A" (ctx.Arg "count") (ctx.Arg "range")
            // todo: ranges are not implemented, because ctx.Arg("range") is always empty (read below)
            [ for i in 1..ctx.Arg("count") -> System.Random.Shared.Next() ]

        Define.Field ("random", ListOf IntType, "Render random numbers", inputs, renderRandoms)
```
and plugged it into existing Query:
```fsharp
    let Query =
        let inputs = [ Define.Input ("id", StringType) ]
        Define.Object<Root> (
            name = "Query",
            fields =
                [ Define.Field ("hero", Nullable HumanType, "Gets human hero", inputs, fun ctx _ -> getHuman (ctx.Arg ("id")))
                  Define.Field ("droid", Nullable DroidType, "Gets droid", inputs, (fun ctx _ -> getDroid (ctx.Arg ("id"))))
                  Define.Field ("planet", Nullable PlanetType, "Gets planet", inputs, fun ctx _ -> getPlanet (ctx.Arg ("id")))
                  Define.Field ("characters", ListOf CharacterType, "Gets characters", (fun _ _ -> characters))
                  randoms // <-- here
                ])
```
Now I can say
```graphql
query { random(count:5) }
```
and get
```json
{ "documentId": -804090008,
  "data": {
    "random": [1773619870, 512090916, 1972640319, 301711584, 480788316]
}}
```
----

### Problem

When I call
```graphql
query { random(count:5, range:{min:10, max:20}) }
```
The `ctx.args("range")` is always passed as an empty unpopulated `obj()` without any type or properties,  
whereas `ctx.args("count")` works as expected.

xperiandri:
That happens because you need to declare
``` F#
type Range = { Min: int; Max: int }
let inputs = [
    Define.Input ("count", IntType)
        Define.Input ("range", Nullable (Define.InputObject<Range>("RandomRange", [
            Define.Input ("min", IntType)
[...]

PR #475, opened 2024-04-06
[RFC] Add support for generic IDictionary<string,obj> query inputs - for #472

pkese:
This will check if InputObject's type is castable to `IDictionary<string, obj>` and will populate that type.

It will work with both `Define.InputObject<Dictionary<string,obj>>` as well as `Define.InputObject<Dynamic.ExpandoObject>`

Idealy we'd reverse the order and first check if we can get a type constructor with matching parameters for a custom type and only then try if the type is castable to `IDictionary<string,obj>` (in case user had defined interface IDictionary on their own custom type).

### DIFF
--- a/src/FSharp.Data.GraphQL.Server/Values.fs
+++ b/src/FSharp.Data.GraphQL.Server/Values.fs
@@ -112,14 +112,19 @@ let rec internal compileByType
                         { new Reflection.ParameterInfo() with
                             member _.Name = f.Name
                             member _.ParameterType = f.TypeDef.Type
-                            member _.Attributes = Reflection.ParameterAttributes.Optional
+                            member _.Attributes =
+                                match f.TypeDef with
+                                | Nullable _ -> Reflection.ParameterAttributes.Optional
+                                | _ -> Reflection.ParameterAttributes.None
                         }
                 |]
                 let constructor (args:obj[]) =
                     let o = Activator.CreateInstance(objtype)
                     let dict = o :?> IDictionary<string, obj>
                     for fld,arg in Seq.zip objDef.Fields args do
-                        dict.Add(fld.Name, arg)
+                        match arg, fld.TypeDef with
+                        | null, Nullable _ -> () // skip populating Nullable fields with nulls
+                        | _, _ -> dict.Add(fld.Name, arg)
                     box o
                 constructor, parameterInfos
             else
--- a/src/FSharp.Data.GraphQL.Server/Values.fs
+++ b/src/FSharp.Data.GraphQL.Server/Values.fs
@@ -105,10 +105,30 @@ let rec internal compileByType
 
     | InputObject objDef ->
         let objtype = objDef.Type
-        let ctor = ReflectionHelper.matchConstructor objtype (objDef.Fields |> Array.map (fun x -> x.Name))
+        let (constructor : obj[] -> obj), (parameterInfos : Reflection.ParameterInfo[]) =
+            if typeof<IDictionary<string,obj>>.IsAssignableFrom(objtype) then
+                let parameterInfos = [|
+                    for f in objDef.Fields ->
+                        { new Reflection.ParameterInfo() with
+                            member _.Name = f.Name
+                            member _.ParameterType = f.TypeDef.Type
+                            member _.Attributes = Reflection.ParameterAttributes.Optional
+                        }
+                |]
+                let constructor (args:obj[]) =
+                    let o = Activator.CreateInstance(objtype)
+                    let dict = o :?> IDictionary<string, obj>
+                    for fld,arg in Seq.zip objDef.Fields args do
+                        dict.Add(fld.Name, arg)
+                    box o
+                constructor, parameterInfos
+            else
+                let ctor = ReflectionHelper.matchConstructor objtype (objDef.Fields |> Array.map (fun x -> x.Name))
+                ctor.Invoke, ctor.GetParameters()
+
 
         let struct (mapper, nullableMismatchParameters, missingParameters) =
-            ctor.GetParameters ()
+            parameterInfos
             |> Array.fold
                 (fun struct (all : ResizeArray<_>, areNullable : HashSet<_>, missing : HashSet<_>) param ->
                     match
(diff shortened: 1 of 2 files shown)


src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -89,10 +89,35 @@ let rec internal compileByType (inputObjectPath: FieldPath) (inputSource : Input
 
     | InputObject objDef ->
         let objtype = objDef.Type
-        let ctor = ReflectionHelper.matchConstructor objtype (objDef.Fields |> Array.map (fun x -> x.Name))
+        let (constructor : obj[] -> obj), (parameterInfos : Reflection.ParameterInfo[]) =
+            if typeof<IDictionary<string,obj>>.IsAssignableFrom(objtype) then
```
xperiandri, line 109: I would also check for `IReadOnlyDictionary<string, obj>` and `obj` and create `ImmutableDictionary<string, obj>` in that case

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -89,10 +89,35 @@ let rec internal compileByType (inputObjectPath: FieldPath) (inputSource : Input
...
+                        { new Reflection.ParameterInfo() with
+                            member _.Name = f.Name
+                            member _.ParameterType = f.TypeDef.Type
+                            member _.Attributes =
+                                match f.TypeDef with
+                                | Nullable _ -> Reflection.ParameterAttributes.Optional
+                                | _ -> Reflection.ParameterAttributes.None
+                        }
+                |]
+                let constructor (args:obj[]) =
```
xperiandri, line 121: And format according to the Fantomas settings
```suggestion
                let constructor (args : obj[]) =
```

xperiandri (commented):
Could you also add some tests to https://github.com/fsprojects/FSharp.Data.GraphQL/tree/dev/tests/FSharp.Data.GraphQL.Tests/Variables%20and%20Inputs ?

pkese:
I think we are dealing here with a bigger problem than just these few cosmetic changes.

The solution presented here is rather hackish. I'm not familiar enough with the codebase to do proper refactoring required to implement this correctly, so I'm fiddling with the code just enough to keep my own app working and afloat.

I think that the real culprit is probably somewhere else, e.g. assuming that input objects are expected to be flat objects where each property is a normal F# type rather than accounting for the case where someone might specify deeply nested input objects.

And providing workarounds at the time of object construction is probably not the right place to fix the issue.

The last commit (https://github.com/fsprojects/FSharp.Data.GraphQL/pull/475/commits/b29ff4b0dda7661aeb711eb919fb8658a80f8077) is a hack around the fact that deeply nested InputsObjects lack `ExecuteInput` member ... again hinting at the fact that the data structures that we built when analyzing input objects specs and its types do not match the stuff that we're trying to parse at runtime.

I guess that this is beyond me and that someone who is a bit more familiar with the code would be able to provide a much cleaner solution to this.

  

Btw,   
this will also break the case if someone would provide a type that implements a `IDictionary` interface (e.g. for https://github.com/fsprojects/FSharp.Data.GraphQL/issues/472#issuecomment-2041054066)
```fsharp
type Range = { Min: int; Max: int }
  with interface IDictionary<string,obj> with
      <<IDictionary implementation..>>
```
because in such case, rather than constructing the Range, it would try to construct and populate the dictionary.
This is fixable, but at the same time the fix would introduce more changes that are probably not going  into the right direction. Current hack works for my personal use, because in my case, all InputObjects are dynamically generated, so I'm not hitting this issue, but for someone else, this may not lead to expected results.

xperiandri:
Thanks, I cap pick this up.
@pkese tell me, please, what was the case for this code?
https://github.com/fsprojects/FSharp.Data.GraphQL/commit/b29ff4b0dda7661aeb711eb919fb8658a80f8077

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -89,10 +89,35 @@ let rec internal compileByType (inputObjectPath: FieldPath) (inputSource : Input
...
+                    let o = Activator.CreateInstance(objtype)
+                    let dict = o :?> IDictionary<string, obj>
+                    for fld,arg in Seq.zip objDef.Fields args do
+                        match arg, fld.TypeDef with
+                        | null, Nullable _ -> () // skip populating Nullable fields with nulls
```
xperiandri, line 126: Why do you want this to happen? If null came and if no filed passed at all are different cases

pkese:
@xperiandri  
The thing was that when defining generic (untyped) objects as inputs, the library will only parse and populate the fist level of inputs.  
If input object contains nested objects, these ended up unpopulated - i.e. `ctx.Args` would have just the top layer of data populated, but not any of the nested inputs (nested inputs in `ctx.Args` would end up being empty `obj()` without any properties).

I assume it is that `ExecuteInput` is not created recursively for subfields, when analyzing `Define.Object<_>`.

pkese:
... so what I did here was that I just mocked what ExecuteInput would have done if it had been defined (here I was assuming it's generic objects all the way down, which in my case it is).

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -163,14 +188,27 @@ let rec internal compileByType (inputObjectPath: FieldPath) (inputSource : Input
                             | ValueSome field ->
                                 match Map.tryFind field.Name props with
                                 | None -> Ok <| wrapOptionalNone param.ParameterType field.TypeDef.Type
+                                | Some input when isNull (box field.ExecuteInput) ->
+                                    // hack around the case where field.ExecuteInput is null
```
Thorium: maybe take this hack as a separate function and not a nested function inside many loops...

xperiandri:
@pkese why would you like overriding Execute?

xperiandri:
Could you reset your branch pointer one commit back to remove unnecessary commit and add a sample of directive usage which showcases the functionality implemented in this PR?

PR #488, opened 2024-09-01
Authorization middleware

xperiandri:


xperiandri:
@njlr this is welcomed to review too

samples/star-wars-api/Startup.fs
```diff
@@ -27,6 +29,12 @@ type Startup private () =
 
     member _.ConfigureServices (services : IServiceCollection) =
         services
+            .AddAuthorization(fun options ->
+                options.AddPolicy (
+                    Policies.CanSetMoon,
+                    (fun policy -> policy.Requirements.Add (IsCharacterRequierment (List.singleton "droid"))))
```
valbers:
```suggestion
                    (fun policy -> policy.Requirements.Add (IsCharacterRequirement (List.singleton "droid"))))
```

samples/star-wars-api/Schema.fs
```diff
@@ -277,24 +267,36 @@ module Schema =
...
+                    ctx.Schema.LiveFieldSubscriptionProvider.Publish<Planet> "Planet" "isMoon" planet
+                    return planet
+                }
+                Define.Field(
+                    "setMoon",
+                    Nullable PlanetType,
+                    "Defines if a planet is actually a moon or not.",
+                    [ Define.Input ("id", StringType); Define.Input ("isMoon", BooleanType) ],
+                    setMoon
+                    // Using complex lambda crashes
```
valbers, line 290: interesting

samples/star-wars-api/Root.fs
```diff
@@ -0,0 +1,12 @@
+namespace FSharp.Data.GraphQL.Samples.StarWarsApi
+
+open System
+open Microsoft.AspNetCore.Http
+open Microsoft.Extensions.DependencyInjection
+
+type Root(ctx : HttpContext) =
+
+    member _.RequestId = ctx.TraceIdentifier
```
valbers, line 9: nice!

samples/star-wars-api/Policies.fs
```diff
@@ -0,0 +1,26 @@
+namespace FSharp.Data.GraphQL.Samples.StarWarsApi.Authorization
+
+open FSharp.Core
+open Microsoft.AspNetCore.Authorization
+
+module Policies =
+
+    let [<Literal>] CanSetMoon = "CanSetMoon"
+
+type IsCharacterRequierment (character : string list) =
```
valbers:
```suggestion
type IsCharacterRequirement (character : string list) =
```

samples/star-wars-api/AuthorizationMiddleware.fs
```diff
@@ -0,0 +1,118 @@
...
+
+    [<Literal>]
+    let AuthorizationPolicy = "AuthorizationPolicy"
+
+    type FieldDef<'Val, 'Res> with
+
+        member this.WithPolicyMiddleware<'Val, 'Res> (middleware : FieldPolicyMiddleware<'Val, 'Res>) : FieldDef<'Val, 'Res> =
+            upcast CustomPolicyFieldDefinition (this, middleware)
+
+        member field.WithAuthorizationPolicies<'Val, 'Res> ([<ParamArray>] policies : string array) : FieldDef<'Val, 'Res> =
```
valbers, line 109: For consistency
```suggestion
        member this.WithAuthorizationPolicies<'Val, 'Res> ([<ParamArray>] policies : string array) : FieldDef<'Val, 'Res> =
```

samples/star-wars-api/AuthorizationMiddleware.fs
```diff
@@ -0,0 +1,118 @@
...
+
+            let middleware ctx value (resolver : ResolveFieldContext -> 'Val -> Async<'Res>) : Async<'Res> = async {
+                let! result = handlePolicies policies ctx value
+                match result with
+                | Ok _ -> return! resolver ctx value
+                | Error error -> return raise (GQLMessageException error)
+            }
+
+            field.WithPolicyMiddleware<'Val, 'Res> middleware
```
valbers, line 118:
```suggestion
            this.WithPolicyMiddleware<'Val, 'Res> middleware
```

samples/star-wars-api/AuthorizationMiddleware.fs
```diff
@@ -0,0 +1,118 @@
...
+        let principal = serviceProvider.GetRequiredService<IHttpContextAccessor>().HttpContext.User
+
+        let! authorizationResults =
+            policies
+            |> Seq.map (fun p -> authorizationService.AuthorizeAsync (principal, value, p))
+            |> Seq.toArray
+            |> Task.WhenAll
+            |> Async.AwaitTask
+
+        let requirements =
```
valbers:
```suggestion
        let failedRequirements =
```

samples/star-wars-api/AuthorizationMiddleware.fs
```diff
@@ -0,0 +1,118 @@
...
+            authorizationResults
+            |> Seq.where (fun r -> not r.Succeeded)
+            |> Seq.collect (fun r -> r.Failure.FailedRequirements)
+
+        if Seq.isEmpty requirements then
```
valbers:
```suggestion
        if Seq.isEmpty failedRequirements then
```

samples/star-wars-api/Policies.fs
```diff
@@ -0,0 +1,26 @@
...
+    member val Characters = character
+    interface IAuthorizationRequirement
+
+type IsCharacterHandler () =
+
+    inherit AuthorizationHandler<IsCharacterRequierment> () // Inject services from DI
+
+        override _.HandleRequirementAsync (context, requirement) =
+            Async.StartAsTask(async {
```
Thorium: If you do want to queue some future work on multi-processor machine, `Async.StartAsTask` is good, it says "hi thread pool, here is a work for you when you are available". But if you want something to execute as fast as possible right now, then consider using `Async.StartImmediateAsTask` which is using current thread instead of thread pool.

samples/star-wars-api/Policies.fs
```diff
@@ -0,0 +1,26 @@
...
+
+open FSharp.Core
+open Microsoft.AspNetCore.Authorization
+
+module Policies =
+
+    let [<Literal>] CanSetMoon = "CanSetMoon"
+
+type IsCharacterRequierment (character : string list) =
+    member val Characters = character
```
Thorium, line 23: Because this Characters is used for searching (List.contains below) then it could be beneficial to have the property as Set instead of List (e.g. just `|> Set.ofList` here). To instantiate a Set is a bit slower than a list, but then when you do `.contains` (like you do in a loop below) then Set is faster for searching.

Thorium (approved):

samples/star-wars-api/Policies.fs
xperiandri: I've changed that in all the places. Could you have a look?

samples/star-wars-api/Policies.fs
xperiandri: Fixed, thanks!

njlr:
> @njlr this is welcomed to review too

Is there a design for this? I always put auth in the resolvers to benefit from railway programming etc.

xperiandri:
The built-in ASP.NET Core functionality is actually very sophisticated as you can see from https://github.com/blowdart/AspNetAuthorizationWorkshop
You can inject IAuthorization service and authorize particular user for particular resource

xperiandri:
This PR adds a bit of shortcut to apply authorization to a fields similar to the way you apply `Authorize` attribute to MVC action or Minimal API handler

xperiandri:
I thougth about this implementation for around 3 years (since the initial prototype)

valbers:
> I thougth about this implementation for around 3 years (since the initial prototype)

Time flies.

@xperiandri Did you see my suggestions?

valbers (changes requested):

xperiandri:
@valbers thank you very much for the suggestions. All applied!

PR #489, opened 2024-09-02
Tiny code clean-up

Thorium:
Tiny code clean-up:

- Avoid dictionary double lookups
- Use isNull instead of "= null"

However, this doesn't affect the execution time performance or memory consumption on any significant manner, because the code-changes are basically on design-time files.


xperiandri (approved):
Thanks so much!

PR #490, opened 2024-09-05
Check if input object CLR property type does not match the scalar's CLR type declared in GraphQL object definition

xperiandri:


src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -140,6 +140,28 @@ module internal ReflectionHelper =
...
+            if ``to``.FullName.StartsWith OptionTypeName || ``to``.FullName.StartsWith ValueOptionTypeName then
+                ``to``.GetGenericArguments().[0]
+            else ``to``
+
+        let result = from.IsAssignableTo ``to``
+        if result then result
+        else
+            if from.FullName.StartsWith OptionTypeName || from.FullName.StartsWith ValueOptionTypeName then
+                from.GetGenericArguments().[0].IsAssignableTo ``to``
+            else result
```
xperiandri: When we declare scalar with a validator like this
``` F#
type AddressLine1 = ValidString<Phantom.Address.Line1>

module Address =
    open ValidString
    open Validus.Operators

    let createLine1 : Validator<string, AddressLine1 voption> =
        (allowEmpty ?=> (Check.String.lessThanLen 1000 <+> validateStringCharacters)) *|* ValueOption.map ValidString

let Line1Type : ScalarDefinition<string, AddressLine1 voption> = Define.ValidStringScalar<AddressLine1>("AddressLine1", createLine1, "Address line 1")
```
it appears that the type of scalar becomes `voption<AddressLine1>`
and then if we declare it `Nullable`
``` F#
Define.Input("line1", Nullable Address.Line1Type)
```
it is double wrapped like `option<voption<AddressLine1>>`
The unwrapping itself happes on execution well but in order to perform check here I need to try unwrapping it 2 times

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -122,27 +131,41 @@ let rec internal compileByType
                             && field.DefaultValue.IsNone
                             ->
                             areNullable.Add param.Name |> ignore
-                        | _ -> all.Add (struct (ValueSome field, param)) |> ignore
+                        | inputDef ->
+                            if ReflectionHelper.isAssignableWithUnwrap inputDef.Type param.ParameterType then
+                                all.Add (struct (ValueSome field, param)) |> ignore
+                            else
+                                // TODO: Consider improving by specifying type mismatches
+                                mismatch.Add param.Name |> ignore
```
xperiandri, line 139: Now we also collect type mismatches

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -122,27 +131,41 @@ let rec internal compileByType
...
+            if typeMismatchParameters.Any () then
+                InvalidInputTypeException (
+                    $"Input object %s{objDef.Name} refers to type '%O{objtype}', but fields '%A{typeMismatchParameters}' have different types than constructor parameters",
+                    typeMismatchParameters.ToImmutableHashSet ()
+                )
+        ]
+        match exceptions with
+        | [] -> ()
+        | [ ex ] -> raise ex
+        | _ -> raise (AggregateException ($"Invalid input object '%O{objtype}'", exceptions))
```
xperiandri, line 168: And throw aggregate exception if multiple issues appear in an input object definition

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -140,6 +144,50 @@ module internal ReflectionHelper =
...
+            ty.GetGenericArguments().[0]
+        else ty
+
+    let isAssignableWithUnwrap (from: Type) (``to``: Type) =
+
+        let checkCollections (from: Type) (``to``: Type) =
+            if
+                // TODO: Implement support of other types of collections using collection initializers
+                (``to``.FullName.StartsWith ListTypeName || ``to``.FullName.StartsWith ArrayTypeName)
+                && (from.IsGenericType
```
Thorium, line 158: Should you also check `to.IsGenericType` because you use it below as taking the [0] argument. Similarly in unwrapOptions above.... it's hard to find the right balance of "how much type-checks will take time" vs "how the code works on edge-cases (when misused by a later PR)"

xperiandri: Yep, before this PR @dim-37 just implemented some corner case I was not thought about 🙂
As I checked that it is `List'1` I suppose I already know it is generic

src/FSharp.Data.GraphQL.Server/Values.fs
```diff
@@ -36,7 +36,8 @@ let private normalizeOptional (outputType : Type) value =
     | value ->
         let inputType = value.GetType ()
         if inputType.Name <> outputType.Name then
-            let expectedOutputType = outputType.GenericTypeArguments[0]
+            // Use only when option or voption so must not be null
+            let expectedOutputType = outputType.GenericTypeArguments.FirstOrDefault()
```
valbers, line 40: If you're sure it will never be `null`, then:
```suggestion
            let expectedOutputType = outputType.GenericTypeArguments.First()
```
Otherwise, I think it's a better idea to actively throw an exception in case of `Default`:
```F#
if expectedOutputType = null
then
   failwith "Expected output type was null. This is a bug in FSharp.Data.GraphQL."
```

xperiandri:
@valbers what do you think?

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -140,6 +144,50 @@ module internal ReflectionHelper =
...
     let isPrameterMandatory = not << isParameterOptional
 
+    let unwrapOptions (ty : Type) =
+        if ty.FullName.StartsWith OptionTypeName || ty.FullName.StartsWith ValueOptionTypeName then
+            ty.GetGenericArguments().[0]
+        else ty
+
+    let isAssignableWithUnwrap (from: Type) (``to``: Type) =
+
+        let checkCollections (from: Type) (``to``: Type) =
```
valbers, line 154: This function looks good. Just some thoughts on coding style: what do you think about getting rid of the `if` altogether and have everything in a single boolean expression?
I mean, remove the `if` and then replace the part:
```F#
then
   let fromType = from.GetGenericArguments()[0]
   let toType = ``to``.GetGenericArguments()[0]
   fromType.IsAssignableTo toType
else
   false
```
with
```F#
&& (from.GetGenericArguments()[0]).IsAssignableTo(``to``.GetGenericArguments()[0])
```
?
The function maybe could also be called `canConvertCollection`

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -140,6 +144,50 @@ module internal ReflectionHelper =
...
+                    )
+
+            then
+                let fromType = from.GetGenericArguments()[0]
+                let toType = ``to``.GetGenericArguments()[0]
+                fromType.IsAssignableTo toType
+            else
+                false
+
+        let from =
```
valbers: another one regarding readability: can we name these `from` and `to` differently? For example: `finalFrom`, `finalTo`?

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
```diff
@@ -140,6 +144,50 @@ module internal ReflectionHelper =
...
+                ``to``.GetGenericArguments()[0]
+            else ``to``
+
+        let result = from.IsAssignableTo ``to`` || checkCollections from ``to``
+        if result then result
+        else
+            if from.FullName.StartsWith OptionTypeName || from.FullName.StartsWith ValueOptionTypeName then
+                let from = from.GetGenericArguments()[0]
+                from.IsAssignableTo ``to`` || checkCollections from ``to``
+            else result
```
valbers:
```suggestion
        from.IsAssignableTo ``to`` || checkCollections from ``to`` ||
        ((from.FullName.StartsWith OptionTypeName || from.FullName.StartsWith ValueOptionTypeName) && ((from.GetGenericArguments()[0]).IsAssignableTo ``to`` || checkCollections from ``to``))
```
also notice how the `else result` in your code is superfluous, since `result` is always false at that point.

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/InputScalarAndAutoFieldScalarTests.fs
```diff
@@ -214,3 +215,55 @@ let ``Execute handles nullable auto-fields in input and output object fields coe
...
+and ApplicationTenantId = ValidString<InputRecord>
+and WrongInput = {
+    Id : ApplicationTenantId
+    /// Legal entity name
+    Name : string
+    LoginProvider : ConsoleLoginProvider
+    AllowedDomains : string list
+    /// Tenants visible to a management company
+    AuthorizedTenants : ApplicationTenantId list
+}
```
valbers, line 235: Interesting. But did you forget to use these types in the test or did you initially intended to use them but then you didn't have to and you forgot them here?

valbers (changes requested):
> @valbers what do you think?

It looks overall good. Please just take a look at my comments.

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri: No, my point was that the `expectedOutputType` value is used only in if clauses that a relevant for `Option` and `ValueOption` types

tests/FSharp.Data.GraphQL.Tests/Variables and Inputs/InputScalarAndAutoFieldScalarTests.fs
xperiandri: You are right! It must be on line 246

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
xperiandri: I thought my version is more readable

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
xperiandri: Good point

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
xperiandri: It is hard to read

src/FSharp.Data.GraphQL.Server/Values.fs
valbers: Ok. Just keep in mind that by doing this you create a logical dependency in the lines of code and someday someone might introduce a bug by not considering your initial intention with `Option` and `ValueOption`.

src/FSharp.Data.GraphQL.Server/ReflectionHelper.fs
valbers: ok, but did you also read the part:
 > also notice how the else result in your code is superfluous, since result is always false at that point.
 
 ?

xperiandri: Yep but do you see wrong in that?

src/FSharp.Data.GraphQL.Server/Values.fs
xperiandri: I have no quick idea how to improve that 🙂

PR #504, opened 2024-11-21
refact: moving websockets code to .Shared to prepare for client impl.

valbers:
We intend to implement the client part now. It's useful to share contract-related code between client and server implementations.


src/FSharp.Data.GraphQL.Shared/Serialization/JsonConverters.fs
```diff
@@ -18,7 +19,7 @@ type ClientMessageConverter () =
     let invalidMsg (explanation : string) = InvalidMessage (4400, explanation) |> Result.Error
 
     let errMsgToStr (struct (docId : int, graphQLErrorMsgs : GQLProblemDetails list)) =
-        String.Join ('\n', graphQLErrorMsgs |> Seq.map (fun err -> err.Message))
+        String.Join ("n", graphQLErrorMsgs |> Seq.map (fun err -> err.Message))
```
valbers: Apparently there's no `Join` overload for using a character as separator when targeting .netstandard2.0.

src/FSharp.Data.GraphQL.Server.AspNetCore/GraphQLSubscriptionsManagement.fs
```diff
@@ -1,6 +1,6 @@
 module internal FSharp.Data.GraphQL.Server.AspNetCore.GraphQLSubscriptionsManagement
 
-open FSharp.Data.GraphQL.Server.AspNetCore.WebSockets
+open FSharp.Data.GraphQL.Shared.Websockets
```
xperiandri:
```suggestion
open FSharp.Data.GraphQL.Shared.WebSockets
```

src/FSharp.Data.GraphQL.Shared/Output.fs
```diff
@@ -0,0 +1,5 @@
+namespace FSharp.Data.GraphQL.Shared
+
+open System.Collections.Generic
+
+type Output = IDictionary<string, obj>
```
xperiandri: I would put `NameValueLookup` into this file

src/FSharp.Data.GraphQL.Shared/Websockets/Exceptions.fs
```diff
@@ -0,0 +1,4 @@
+namespace FSharp.Data.GraphQL.Shared.Websockets
```
xperiandri:
```suggestion
namespace FSharp.Data.GraphQL.Shared.WebSockets
```

src/FSharp.Data.GraphQL.Shared/Websockets/Messages.fs
```diff
@@ -1,10 +1,9 @@
-namespace FSharp.Data.GraphQL.Server.AspNetCore.WebSockets
+namespace FSharp.Data.GraphQL.Shared.Websockets
```
xperiandri:
```suggestion
namespace FSharp.Data.GraphQL.Shared.WebSockets
```

src/FSharp.Data.GraphQL.Shared/Websockets/Messages.fs
xperiandri, line 1: I would put `InvalidWebsocketMessageException` into this file and name it `WebSockets.fs`

src/FSharp.Data.GraphQL.Shared/Output.fs
valbers: Don't you think it's kind of confusing to name the file `Output.fs` and there's just one line of code defining `Output` as an alias for `Dictionary<string, obj>` while the rest of the file, which is not small by the way, is all just the definition of `NameValueLookup` 🙂 ?

src/FSharp.Data.GraphQL.Server.AspNetCore/GraphQLSubscriptionsManagement.fs
valbers: Done.

src/FSharp.Data.GraphQL.Shared/Websockets/Messages.fs
valbers: Done.

src/FSharp.Data.GraphQL.Shared/Websockets/Messages.fs
valbers: Done.

src/FSharp.Data.GraphQL.Shared/Websockets/Exceptions.fs
valbers: Done.

src/FSharp.Data.GraphQL.Shared/Serialization/JsonConverters.fs
xperiandri: Ah, yes. However the slash disappeared
```suggestion
        String.Join ("\n", graphQLErrorMsgs |> Seq.map (fun err -> err.Message))
```

xperiandri (changes requested):

src/FSharp.Data.GraphQL.Shared/Output.fs
xperiandri: I think that is fine. As `NameValueLookup` is an `Output`

xperiandri: Or we can go opposide way and move this interface there

valbers: > As NameValueLookup is an Output

ok. I moved `NameValueLookup` to `Output.fs`.

samples/chat-app/server/Program.fs
```diff
@@ -29,7 +29,6 @@ let main args =
     if app.Environment.IsDevelopment () then
         app.UseGraphQLAltair "/altair" |> ignore
         app.UseGraphQLGraphiQL "/graphiql" |> ignore
-        app.UseGraphQLPlayground "/playground" |> ignore
```
xperiandri, line 32: The package is not deprecated, maybe just a method is deprecated?

valbers: Now that you pointed that out, I just saw [that the whole package `GraphQL.Server.Ui.Playground` is deprecated](https://github.com/graphql-dotnet/server). Using GraphQLPlayground here was [causing the builds to fail](https://github.com/fsprojects/FSharp.Data.GraphQL/actions/runs/12153973380).

xperiandri: Ah, I see. The benefit of the Playground is the ability to get schema in DSL format.
If it is deprecated, then let's remove it

xperiandri: Then we also need to remove a line in `Packages.props`

xperiandri (approved):
Looks good. If no more changes In tended I will merge

PR #507, opened 2024-12-07
Double coercion and error messages fixes

xperiandri:
1. Because coercion collections were of type `seq` the error-splitting logic called them twice causing double execution
2. Because collections passed to input object fields mismatch validation were not F# collections their string values used to not appear in the exception message.


valbers (approved):
LGTM

PR #508, opened 2024-12-08
Implemented the ability to recognize if an input field is null or not present at all

xperiandri:
Imagine you have an address
``` F#
type AddressRecord = {
    Line1: AddressLine1 Skippable
    Line2: AddressLine2 voption Skippable
    City: City Skippable
    State: State Skippable
    ZipCode: ZipCode Skippable
}
```
and you want to patch it (apply changes to only particular fields keeping the other untouched).

Then you can send such an input object
``` JSON
{
  "line1": "New value"
  "line2": null // means clear value
  // skip other properties, means do not change them
}
```

To support this approach we need to be able to recognize if a field is null or missing and to support converting input values into `System.Text.Json.Serialization.Skippable` where missing value is `Skip`, `null` value is `Include None`/`Include ValueNone` (when class constructor parameter is `Option`/`ValueOption` and `Skip` otherwise) and `value` is `Include value`

See StarWarsApi for usage sample

### DIFF
--- a/samples/star-wars-api/Schema.fs
+++ b/samples/star-wars-api/Schema.fs
@@ -1,5 +1,6 @@
 namespace FSharp.Data.GraphQL.Samples.StarWarsApi
 
+open System.Text.Json.Serialization
 open FSharp.Data.GraphQL
 open FSharp.Data.GraphQL.Types
 open FSharp.Data.GraphQL.Server.Relay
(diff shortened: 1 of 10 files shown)


xperiandri:
@njlr @pkese what do you think?

njlr:
> @njlr @pkese what do you think?

LGTM 👍

pkese:
Yes, it's a good solution. :+1:

PR #528, opened 2025-05-05
Fixed `graphql-transport-ws` implementation

valbers:
Here I fix the _graphql-transport-ws_ implementation.
* The payload format in `Next` messages was wrong: it was not the standard `{ "data": {}, "errors": [] }` (i.e.: `ExecutionResult`) but just the data object directly.
* `Error` is now only sent as a response to a `Subscribe` message and not elsewhere;
* Fixed bug of our default websockets endpoint `/ws` path not being set (not related to the _graphql-transport-ws_ subprotocol, but anyway to websockets configuration).

### DIFF
--- a/src/FSharp.Data.GraphQL.Shared/WebSockets.fs
+++ b/src/FSharp.Data.GraphQL.Shared/WebSockets.fs
@@ -3,6 +3,7 @@ namespace FSharp.Data.GraphQL.Shared.WebSockets
 open System
 open System.Collections.Generic
 open System.Text.Json
+open FSharp.Data.GraphQL
 open FSharp.Data.GraphQL.Shared
 
 type InvalidWebsocketMessageException (explanation : string) =
@@ -15,8 +16,10 @@ type SubscriptionsDict = IDictionary<SubscriptionId, SubscriptionUnsubscriber * 
 
 type RawMessage = { Id : string voption; Type : string; Payload : JsonDocument voption }
 
+type SubscriptionExecutionResult = { Data : Output voption; Errors : GQLProblemDetails list }
+
 type ServerRawPayload =
-    | ExecutionResult of Output
+    | ExecutionResult of SubscriptionExecutionResult
     | ErrorMessages of NameValueLookup list
     | CustomResponse of JsonDocument
 
@@ -35,7 +38,7 @@ type ServerMessage =
     | ConnectionAck
     | ServerPing
     | ServerPong of JsonDocument voption
-    | Next of id : string * payload : Output
+    | Next of id : string * payload : SubscriptionExecutionResult
     | Error of id : string * err : NameValueLookup list
     | Complete of id : string
 
--- a/src/FSharp.Data.GraphQL.Server.AspNetCore/GraphQLWebsocketMiddleware.fs
+++ b/src/FSharp.Data.GraphQL.Server.AspNetCore/GraphQLWebsocketMiddleware.fs
@@ -169,33 +169,27 @@ type GraphQLWebSocketMiddleware<'Root>
         let sendMsg = sendMessageViaSocket serializerOptions socket
         let rcv () = socket |> rcvMsgViaSocket serializerOptions
 
-        let sendOutput id (output : Output) =
-            match output.TryGetValue "errors" with
-            | true, theValue ->
-                // The specification says: "This message terminates the operation and no further messages will be sent."
-                subscriptions
-                |> GraphQLSubscriptionsManagement.removeSubscription (id)
-                sendMsg (Error (id, unbox theValue))
-            | false, _ -> sendMsg (Next (id, output))
+        let sendOutput id (output : SubscriptionExecutionResult) =
+            sendMsg (Next (id, output))
 
         let sendSubscriptionResponseOutput id subscriptionResult =
             match subscriptionResult with
-            | SubscriptionResult output -> output |> sendOutput id
+            | SubscriptionResult output -> { Data = ValueSome output; Errors = [] } |> sendOutput id
             | SubscriptionErrors (output, errors) ->
                 logger.LogWarning ("Subscription errors: {subscriptionErrors}", (String.Join ('\n', errors |> Seq.map (fun x -> $"- %s{x.Message}"))))
-                Task.FromResult ()
+                { Data = ValueNone; Errors = errors } |> sendOutput id
 
         let sendDeferredResponseOutput id deferredResult =
             match deferredResult with
             | DeferredResult (obj, path) ->
                 let output = obj :?> Dictionary<string, obj>
-                output |> sendOutput id
+                { Data = ValueSome output; Errors = [] } |> sendOutput id
             | DeferredErrors (obj, errors, _) ->
                 logger.LogWarning (
                     "Deferred response errors: {deferredErrors}",
                     (String.Join ('\n', errors |> Seq.map (fun x -> $"- %s{x.Message}")))
                 )
-                Task.FromResult ()
+                { Data = ValueNone; Errors = errors } |> sendOutput id
 
         let sendDeferredResultDelayedBy (ct : CancellationToken) (ms : int) id deferredResult : Task = task {
             do! Task.Delay (ms, ct)
@@ -208,16 +202,16 @@ type GraphQLWebSocketMiddleware<'Root>
                 (subscriptions, socket, observableOutput, serializerOptions)
                 |> addClientSubscription id sendSubscriptionResponseOutput
             | Deferred (data, errors, observableOutput) ->
-                do! data |> sendOutput id
+                do! { Data = ValueSome data; Errors = [] } |> sendOutput id
                 if errors.IsEmpty then
                     (subscriptions, socket, observableOutput, serializerOptions)
                     |> addClientSubscription id (sendDeferredResultDelayedBy cancellationToken 5000)
                 else
                     ()
-            | Direct (data, _) -> do! data |> sendOutput id
+            | Direct (data, _) -> do! { Data = ValueSome data; Errors = [] } |> sendOutput id
             | RequestError problemDetails ->
                 logger.LogWarning("Request errors:\n{errors}", problemDetails)
-
+                do! { Data = ValueNone; Errors = problemDetails } |> sendOutput id
         }
 
         let logMsgReceivedWithOptionalPayload optionalPayload (msgAsStr : string) =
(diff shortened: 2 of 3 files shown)


src/FSharp.Data.GraphQL.Shared/WebSockets.fs
```diff
@@ -15,8 +16,10 @@ type SubscriptionsDict = IDictionary<SubscriptionId, SubscriptionUnsubscriber *
 
 type RawMessage = { Id : string voption; Type : string; Payload : JsonDocument voption }
 
+type SubscriptionExecutionResult = { Data : Output voption; Errors : GQLProblemDetails list }
```
xperiandri, line 19: As I understand from [this](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#next) it uses the same execution result as the HTTP version of the GraphQL.
Have you ever found there `ExecutionResult` declared in the spec?

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
```diff
@@ -35,7 +38,7 @@ type ServerMessage =
     | ConnectionAck
     | ServerPing
     | ServerPong of JsonDocument voption
-    | Next of id : string * payload : Output
+    | Next of id : string * payload : SubscriptionExecutionResult
     | Error of id : string * err : NameValueLookup list
     | Complete of id : string
```
xperiandri, line 43: Also it would be nice to make these DUs serializable right away using 
`[<JsonFSharpConverter(unionTagName = "type")>]`
This will eliminate the need of `serializeServerMessage` function

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
```diff
@@ -15,8 +16,10 @@ type SubscriptionsDict = IDictionary<SubscriptionId, SubscriptionUnsubscriber *
 
 type RawMessage = { Id : string voption; Type : string; Payload : JsonDocument voption }
```
xperiandri, line 17: Do we use it anywhere?

xperiandri (changes requested):
Have you found a better spec than this one?
https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md
It would be nice to follow the naming and add references to the spec as we do in AST code

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
xperiandri: For each union case you will need to apply attributes like `[<JsonName "connection_ask">]` then

valbers: Are you sure this will make the code better? I think that how it is now it's simpler and maybe more flexible?

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
valbers: It is indeed the same -- as long as the mandatory properties are concerned. But we also define `documentId` additionally, which wouldn't be necessary here, I guess. And, so far, we're not setting `extensions` either.

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
valbers: Yes: in `serializeServerMessage` 🙂

valbers:
> Have you found a better spec than this one? https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md It would be nice to follow the naming and add references to the spec as we do in AST code

References to the spec 👍 .

Well, this is the official specification. As far as I know, this subprotocol is anyway something more like a _de facto_ protocol than a standard. Or what do you mean by "better spec"? In the Apollo GraphQL documentation, for example, you can see a reference to it: https://www.apollographql.com/docs/react/data/subscriptions#supported-subscription-protocols.

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
valbers: By `ExecutionResult` I understand this: https://www.graphql-js.org/api-v16/execution/

xperiandri: I wonder from where `documentId` came 🙂

xperiandri: Then we ought to use
```
interface ExecutionResult<
  TData = ObjMap<unknown>,
  TExtensions = ObjMap<unknown>,
> {
  errors?: ReadonlyArray<GraphQLError>;
  data?: TData | null;
  extensions?: TExtensions;
}
```
in our code.
Then we can share that record between client and server

src/FSharp.Data.GraphQL.Shared/WebSockets.fs
```diff
@@ -13,31 +14,24 @@ type SubscriptionUnsubscriber = IDisposable
...
-type RawMessage = { Id : string voption; Type : string; Payload : JsonDocument voption }
-
-type ServerRawPayload =
-    | ExecutionResult of Output
-    | ErrorMessages of NameValueLookup list
-    | CustomResponse of JsonDocument
-
-type RawServerMessage = { Id : string voption; Type : string; Payload : ServerRawPayload voption }
-
+[<JsonFSharpConverter(unionTagName = "type", SkippableOptionFields = SkippableOptionFields.Always)>]
```
valbers: Granted, this approach uses less code, but the error messages are less helpful, especially for a consumer who is using a whole different platform than dotnet. See the failing tests, for example:
```
           ↓ (pos 0)
      Expected: "Property "id" is required for this messag"···
      Actual:   "Missing field for union type FSharp.Data."···
                 ↑ (pos 0)
 ```
I think this approach (with `JsonFSharpConverter` annotations) is fine for application code, i.e. code you'd use in your own internal application, but since `FSharp.Data.GraphQL` is a library for broader use, I think we should have more customized error messages. So this is the case for us to have our own JSON converter, like in the previous implementation. What do you think?

File: src/FSharp.Data.GraphQL.Shared/WebSockets.fs (hunk 1 above)
Comment (valbers) [reply src/FSharp.Data.GraphQL.Shared/WebSockets.fs:19]: Regarding `documentId`, this is what I found in the [documentation](https://fsprojects.github.io/FSharp.Data.GraphQL/execution-pipeline.html#Planning-phase):
>_Spliting planning and execution phases is a good idea, when you have the same GraphQL query requested many times (with potentially different variables). This way you can compute execution plan once and cache it. You can use **executionPlan.DocumentId** as a cache identifier. **DocumentId** is also returned as one of the top level fields in GraphQL response, so it can be used from the client side. Other GraphQL implementations describe that technique as persistent queries._

Maybe we might want to implement it like in [Apollo GraphQL](https://www.apollographql.com/docs/kotlin/advanced/persisted-queries). Anyway, if we remove `documentId` now, it'd be a breaking change in our library.

File: src/FSharp.Data.GraphQL.Server.AspNetCore/GQLResponseExtensions.fs [hunk 5]
```diff
@@ -0,0 +1,60 @@
+[<AutoOpen>]
+module internal FSharp.Data.GraphQL.GraphQLWebSocketResponseExtensions
+
+open System
+open System.Collections.Generic
+open System.Text.Json.Serialization
+open FSharp.Data.GraphQL
+open FSharp.Data.GraphQL.Extensions
+
+type GQLWebSocketResponse with
```
valbers: Do we really need so many overloads right now? One of the things I like about F# is that you can't overload functions, so it feels like a step backwards to use them via classes in F#. I'd prefer to have a record and fill the record accordingly when instantiating it. It's also more visible to reader. I mean: I prefer reading
```fsharp
{ Data = Include data
  Path = Skip
  HasNext = Skip
  Errors = Skip
  Extensions = Skip }
```
than reading
```fsharp
GQLWebSocketResponse.Direct(data)
```
because then, as a reader, I can see at a glance what the line is actually about and with `GQLWebSocketResponse.Direct(data)` it's hidden away from me, so I have to navigate to the code and maybe get lost among the many overload definitions.

src/FSharp.Data.GraphQL.Server/IO.fs
```diff
@@ -7,24 +7,23 @@ open FSharp.Data.GraphQL
...
-          Data = Include data
-          Errors = Skippable.ofList errors }
-    static member Stream(documentId) =
-        { DocumentId = documentId
-          Data = Include null
-          Errors = Skip }
-    static member RequestError(documentId, errors) =
-        { DocumentId = documentId
-          Data = Skip
-          Errors = Include errors }
```
valbers: What do you think about separating concerns for the sake of pull requests? I mean, this pull request was just supposed to fix the `graphql-transport-ws` implementation, but with this change we're introducing an important breaking change that affects everything else...
I think we should handle those changes in a PR for redesigning the global `GQLResponse` format.

src/FSharp.Data.GraphQL.Shared/Serialization/JSON.fs
```diff
@@ -16,11 +16,7 @@ let configureSerializerOptions (jsonFSharpOptions: JsonFSharpOptions) (additiona
     jsonFSharpOptions.AddToJsonSerializerOptions options
 
 let configureWSSerializerOptions (jsonFSharpOptions: JsonFSharpOptions) (additionalConverters: JsonConverter seq) (options : JsonSerializerOptions) =
-    let additionalConverters = seq {
-        yield new ClientMessageConverter () :> JsonConverter
-        yield new RawServerMessageConverter ()
-        yield! additionalConverters
-    }
+    let jsonFSharpOptions = jsonFSharpOptions.WithSkippableOptionFields (SkippableOptionFields.Always, true)
```
valbers: Can it be the reason why some tests are failing (tests in which the response object now doesn't contain properties that have `null` values anymore)? I don't know how this configuration here for websockets would leak into these seemingly unrelated tests, though...

src/FSharp.Data.GraphQL.Server.AspNetCore/GQLResponseExtensions.fs
xperiandri: I thought it is too much to fill and transform and it adds unnecessary cognitive load in the place of usage

valbers: I get it, but, as said, I think this is one of those things which look good at a first sight, but have higher costs in the long run. Anyway, this is just my opinion.

src/FSharp.Data.GraphQL.Server/IO.fs
xperiandri: OK, let's return your branch to your changes and mine will go next

xperiandri (approved):

PR #583, opened 2026-06-13
NuGet packaging of DesignTime DLL for F# provider

xperiandri:
Added MSBuild target to ensure DesignTime DLL is included in NuGet package, working around F# SDK limitations with ProjectReference-based providers. Removed obsolete target for cleaning NuGet pack input and updated conditions for consistency.

### DIFF
--- a/src/FSharp.Data.GraphQL.Client.DesignTime/FSharp.Data.GraphQL.Client.DesignTime.fsproj
+++ b/src/FSharp.Data.GraphQL.Client.DesignTime/FSharp.Data.GraphQL.Client.DesignTime.fsproj
@@ -19,7 +19,7 @@
       <!-- This excludes FSharp.Core.xml and FSharp.Core.resources.dll while referencing the right FSharp.Core.dll version -->
       <ExcludeAssets>contentFiles;runtime</ExcludeAssets>
     </PackageReference>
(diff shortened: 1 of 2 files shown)


xperiandri:
@T-Gro, I discovered that the design-time DLL is no longer packaged into the NuGet package.
Could you confirm that this solution is the right way?

T-Gro:
@xperiandri : I will take a look and write here 👍

T-Gro:
@xperiandri heads-up: the root cause behind this workaround is being fixed upstream in dotnet/fsharp#19979. `PackageFSharpDesignTimeTools` will again package a provider activated via `<ProjectReference IsFSharpDesignTimeProvider="true">` on its own — resolved through the SDK's nearest-target-framework negotiation, with no reference build (so `pack --no-build` stays fine).

Once that ships in a future .NET SDK, the custom `_IncludeDesignTimeDllInPackage` target here can be removed: the stock gesture packages `FSharp.Data.GraphQL.Client.DesignTime` (and its sibling DLLs) into `typeproviders/fsharp41/<tfm>/`. Keep `<FSharpToolsDirectory>typeproviders</FSharpToolsDirectory>` to retain the current layout. No rush — please wait for the SDK that carries the fix before dropping the workaround.

